PR #9, opened 2017-02-07
Fix for Stream player properties failing to be updated

0xb1dd1e:
The edits weren't being applied to the SerializedPropety.


i-saint:
thanks!

PR #19, opened 2017-04-20
Fixed transform issue with points renderer.

diftow:
https://github.com/unity3d-jp/AlembicImporter/issues/17


i-saint:
thanks, but we can't accept because this is far from enough for our cases.

diftow:
Are you still implementing your own fix then?

i-saint:
I will work on it next week, as I said. (I'm on vacation now)
but I think your fix is enough for your particular case.

PR #43, opened 2018-01-17
Store relative paths in Stream Descriptors, to avoid loading errors in builds

Chosko:
The field "pathToAbc" in AlembicStreamingDescriptor was filled with absolute paths. 
When making a build and deploying it to another PC, there was always an error: "failed to load alembic at...." because the absolute paths of the alembic streaming assets were different.
With these changes we save the relative paths in the field "pathToAbc", and we compute the absolute paths at runtime.


i-saint:
thanks!

PR #78, opened 2018-03-07
Fixes for Linux

Tak:
Fixup module importer for linux
Add ModuleImporter.cs.meta


benoithudson:
Oops, this is obsolete with the work I just merged in. No more ModuleImporter, just DllImport with a messy path, and it works on all three platforms (except that libabci.so will still have ABI issues).

Tak:
Ok cool, I'll check out the new changes tomorrow

PR #96, opened 2018-04-20
Fix Exporter target switching

ErikBehar:
This line of code was changing the target to always be current GameObject instead of whatever you dragged into the public property. Either change the wording to make it clear that you should drop component on the target, or remove this line as proposed. To elaborate:
A) you should not show a public target field on the inspector if you are meant to drop this component on the branch target. 
or
B) take pull request to make the target keep the GameObject that has been dragged into the filed.


clusty:
Closing very old PR.

ErikBehar:
its a very simple change, why close it if its still applicable?

clusty:
Sorry for the cryptic message.  I did not think anybody still cared about it.
This is already fixed  since a while now:)

ErikBehar:
Great ! Thanks

PR #97, opened 2018-04-20
Null Reference if mesh destroyed

ErikBehar:
Was trying to export some meshes that are destroyed randomly, ran into this null reference error, this was fast fix. Not sure if there are negative side effects by fixing this way ?


i-saint:
thanks for your feedback! I will test it.

ErikBehar:
suggest closing this PR ... see PR #100 instead

PR #99, opened 2018-04-24
Uni 45605 Windows CI

inwoods:
@benoithudson I tried and failed to make the cmake build work using the external ThirdParty directory (HDF5,Alembic,ilmbase etc) so I opted for batch script.

I've modified the abci.vcproj to point to this new ThirdParty directory.


Plugin/external/ThirdParty/Deploy/Windows/Alembic/include/Alembic/Abc/All.h
```diff
@@ -0,0 +1,76 @@
+//-*****************************************************************************
```
benoithudson: (a) kind of tragic we can't install all three platforms in the same place (and thus share the .h files)

(b) why is this in Windows/Alembic/include rather than in Windows/include? The cmake assumes you install all the dependences in a common prefix location.

.gitignore
```diff
@@ -61,5 +61,6 @@ StreamingAssets*
 # Unignore specific folders
 !Plugin/external/ThirdParty/Deploy/MacOS/include
 !Plugin/external/ThirdParty/Deploy/MacOS/lib
+!Plugin/external/ThirdParty/Deploy/Windows/
```
benoithudson: Why are we including all of Windows rather than just two directories like on Mac?

.gitlab-ci.yml
```diff
@@ -45,8 +45,12 @@ build:windows:
     - rmdir /s /q build
     - md build
     - cd build
-    #- cmake ../Plugin
```
benoithudson: On mac we needed to pass a bunch of flags here.

.gitlab-ci.yml
```diff
@@ -45,8 +45,12 @@ build:windows:
...
-    #- make
```
benoithudson: I recall using nmake rather than make on windows.

.gitlab-ci.yml
```diff
@@ -45,8 +45,12 @@ build:windows:
...
+    - set builddir %cd%
+    - for /f "usebackq tokens=*" %i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" /latest /property installationPath`) do (set VSDIR=%i)
+    - call "%VSDIR%\Common7\Tools\VsDevCmd.bat"
+    - cd /d %builddir%
+    - msbuild ..\Plugin\abci.sln /t:Build /p:Configuration=Master /p:Platform=x64 /m /nologo /p:OutDir=%builddir%\_out\ /p:IntDir=%builddir%\_tmp\
```
benoithudson: Can you add directions on how to update the abci.sln file to the readme?

Plugin/external/ThirdParty/Deploy/Windows/Alembic/include/Alembic/Abc/All.h
inwoods: I did the easiest thing possible to get the abci.sln building. This is the way External.7z is organized - I'm unsure on the which actual versions of openexr, ilmbase, alembic have been built.

more work needs to be done to make the platform structure common.

.gitignore
inwoods: again - I want to make the minimum change to get abci.sln building. I unpacked [External.7z](https://github.com/unity3d-jp/AlembicForUnity/releases/download/20180122/External.7z) into a single folder.

.gitlab-ci.yml
inwoods: what kind of information are you looking for?

the paths to includes and libs are in the abci.vcproj

PR #100, opened 2018-04-24
Fix new objects parent CaptureNode is null

ErikBehar:
This further fixes PR #97 ( that PR can be closed ).
 
Problem:
When instantiating new prefabs with a hierarchy and adding to target export branch 
if the prefab has a mesh that is child to transform, when the child is added the 
parent capture node will be false.

Fix:
recurse up hierarchy adding new transform nodes if they haven't been added already
when adding new capture nodes to children.

Other Changes :
* warn when this happens
* optimize update loop to only new adds

### DIFF
--- a/AlembicImporter/Assets/UTJ/Alembic/Scripts/Exporter/AlembicExporter_impl.cs
+++ b/AlembicImporter/Assets/UTJ/Alembic/Scripts/Exporter/AlembicExporter_impl.cs
@@ -295,14 +295,14 @@ namespace UTJ.Alembic
             {
                 var target = c as Transform;
                 if (parent == null || target == null)
-				{
-					if (parent == null)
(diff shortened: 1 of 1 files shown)


clusty:
Closing very old PR.

ErikBehar:
Just because its old doesn't mean you shouldnt still fix this.

clusty:
Hey,
Don't worry, it will be fixed. It's already tracked internally.
I'll take the fix over and push it through the proper process: 
This PR being so old, it's a PITA to resolve conflicts since repo went through multiple massive changes that broke history.
Also  being a fork, there is no way to run our  automation suite through it :(
Will keep you posted when it lands if you  wish.

ErikBehar:
Ah great, thanks!

PR #123, opened 2018-06-11
Uni-49877 Proper naming for the Alembic package

lassond:
* Changed the namespace and assembly names
* Removed deprecated code


AlembicImporter/Packages/com.unity.formats.alembic/Editor/Misc/BoolDrawer.cs
```diff
@@ -1,7 +1,8 @@
 using UnityEngine;
+using UnityEngine.Formats.Alembic.Sdk;
 using UnityEditor;
 
-namespace UTJ.Alembic
+namespace UnityEditor.Formats.Alembic.Importer
 {
```
lassond: This could also be in a "Util" namespace but since only the importer uses it I left it there

AlembicImporter/Packages/com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicCameraParams.cs
```diff
@@ -1,6 +1,6 @@
 using UnityEngine;
 
-namespace UTJ.Alembic
+namespace UnityEngine.Formats.Alembic.Util
 {
```
lassond: Typically the Util namespace is used for things that are required from both the importer and the exporter

AlembicImporter/Packages/com.unity.formats.alembic/Runtime/Scripts/Misc/Bool.cs
```diff
@@ -1,7 +1,7 @@
 using System;
 using UnityEngine;
 
-namespace UTJ.Alembic
+namespace UnityEngine.Formats.Alembic.Sdk
 {
```
lassond: The Sdk namespace depends on the Bool class, so I added the Bool class to the Sdk namespace

AlembicImporter/Packages/com.unity.formats.alembic/Runtime/Scripts/Misc/PinnedArray.cs
```diff
@@ -103,6 +103,10 @@ System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
     }
 
 
+}
+
+namespace UnityEngine.Formats.Alembic.Sdk
+{
```
lassond: The PinnedList class is also used in the Sdk namespace, to marshall data to/from c++

AlembicImporter/Packages/com.unity.formats.alembic/Editor/Timeline/AlembicRecorderClipEditor.cs
```diff
@@ -1,12 +1,13 @@
...
 using UnityEngine;
+using UnityEngine.Formats.Alembic.Util;
 using UnityEngine.SceneManagement;
 using UnityEditor;
 using UnityEditor.SceneManagement;
 using UnityEngine.Timeline;
 
-namespace UTJ.Alembic
+namespace UnityEditor.Formats.Alembic.Importer
 {
```
lassond: Do you think we should create a "Timeline" namespace for the timeline bits? I am not too familiar with the package yet but it seems like this is a separate set of features.

vkovec: yes I think that would make more sense than having the timeline stuff under "importer"

AlembicImporter/Assets/Tests/Scripts/ParticleEngine.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.Runtime.InteropServices;
 using UnityEngine;
-
+using UnityEngine.Formats.Alembic.Sdk;
 
 namespace UTJ.Alembic
```
vkovec: should this namespace be changed as well?

AlembicImporter/Assets/Tests/Scripts/ParticleEngineCapturer.cs
```diff
@@ -1,5 +1,7 @@
 using System.Runtime.InteropServices;
 using UnityEngine;
+using UnityEngine.Formats.Alembic.Sdk;
+using UnityEngine.Formats.Alembic.Util;
 
 namespace UTJ.Alembic
```
vkovec: this namespace too

AlembicImporter/Packages/com.unity.formats.alembic/Runtime/Scripts/Misc/PinnedArray.cs
vkovec: should PinnedObject also be in the sdk namespace?

lassond: I don't know. Only PinnedList is used by the Sdk. The other ones (PinnedObject and PinnedArray) are used by higher-level assemblies. 

But I find it would be more elegant to have them all in the same namespace so I will move them

AlembicImporter/Assets/Tests/Scripts/ParticleEngine.cs
lassond: These are manual tests, we don't care I think

PR #124, opened 2018-06-12
Uni 49870 fix FDG code analysis warnings

vkovec:
Warning: these changes break some of the tests. To be fixed in another branch.

TODO: test with the unit tests.


AlembicImporter/Packages/com.unity.formats.alembic/Editor/Misc/BoolDrawer.cs
```diff
@@ -6,8 +6,15 @@ namespace UTJ.Alembic
     [CustomPropertyDrawer(typeof(Bool))]
     class BoolDrawer : PropertyDrawer
     {
+        private BoolDrawer() { }
```
vkovec: warning suggested to add private constructor as class is never instantiated. Will have to make sure this doesn't cause issues

AlembicImporter/Packages/com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
```diff
@@ -22,38 +22,147 @@ public enum ExportScope
...
+        public bool DebugLog
+        {
+            get { return debugLog; }
+            set { debugLog = value; }
+        }
     }
 
     [AttributeUsage(AttributeTargets.Class)]
-    public class CaptureTarget : Attribute
+    public sealed class CaptureTarget : Attribute
```
lassond: Do you remember why you need to seal this class? Just curious...

AlembicImporter/Packages/com.unity.formats.alembic/Runtime/Scripts/Importer/AbcAPI.cs
```diff
@@ -453,86 +414,53 @@ public struct aiSample
...
-        [DllImport(Abci.Lib)] static extern aiPropertyType aiPropertyGetType(IntPtr prop);
-        [DllImport(Abci.Lib)] static extern void aiPropertyGetData(IntPtr prop, aiPropertyData oData);
-        #endregion
-    }
-
-    public partial class AbcAPI
-    {
-        [DllImport(Abci.Lib)] public static extern aiSampleSelector aiTimeToSampleSelector(double time);
-        [DllImport(Abci.Lib)] public static extern void aiCleanup();
+        public static bool ToBool(aiProperty v) { return v; }
```
lassond: What does that do?

AlembicImporter/Packages/com.unity.formats.alembic/Runtime/Scripts/Timeline/AlembicRecorderTrack.cs
```diff
@@ -5,7 +5,6 @@ namespace UTJ.Alembic
 {
     [System.Serializable]
     [TrackClipType(typeof(AlembicRecorderClip))]
-    [TrackMediaType(TimelineAsset.MediaType.Script)]
```
lassond: Thanks for removing the deprecated attributes 👍

AlembicImporter/Packages/com.unity.formats.alembic/Runtime/Scripts/Importer/AbcAPI.cs
vkovec: it's an alternative for `public static implicit operator bool(aiProperty v) { return v.self != IntPtr.Zero; }`

There's a warning for giving people an alternative if they can't use the operator directly for some reason.

lassond: OK, and I guess there it will call the operator right? "return v" becomes "(bool) return v" which calls the operator. Thanks

PR #130, opened 2018-06-20
Uni 46254 mvp alembic ui

vkovec:
- hid timeline recorder + camera params

### DIFF
--- a/AlembicImporter/Packages/com.unity.formats.alembic/Editor/Timeline/AlembicRecorderClipEditor.cs
+++ b/AlembicImporter/Packages/com.unity.formats.alembic/Editor/Timeline/AlembicRecorderClipEditor.cs
@@ -8,6 +8,7 @@ using UnityEditor;
 using UnityEditor.SceneManagement;
 using UnityEngine.Timeline;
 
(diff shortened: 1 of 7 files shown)


AlembicImporter/Packages/com.unity.formats.alembic/Editor/Timeline/AlembicRecorderClipEditor.cs
```diff
@@ -156,3 +157,4 @@ TimelineAsset FindTimelineAsset()
         }
     }
 }
+#endif
```
lassond: Good practice is to add a comment following the endif, indicating what it relates to. Here for example:
#endif // ENABLE_ALEMBIC_TIMELINE_RECORDER

lassond:
Does that cover the "end of frame" component? Is it gone from the component search dialog now?

vkovec:
@lassond yes I think I added the #ifdef to that component as well

vkovec:
@lassond unhiding the timeline recorder per Seiya's request

vkovec:
ran the runtime tests and they all pass with these changes

lassond:
Perfect, please merge!

PR #135, opened 2018-06-22
Uni 51131 add basic unit tests

vkovec:


AlembicImporter/Assets/UnitTests/AlembicExportTest.cs
```diff
@@ -0,0 +1,235 @@
...
+
+            var absPath = GetAssetsAbsolutePath(abcPath);
+
+            Assert.That(string.IsNullOrEmpty(absPath), Is.False);
+            Assert.That(File.Exists(absPath));
+
+            // now try loading the asset to see if it imported properly
+            var obj = AssetDatabase.LoadMainAssetAtPath(abcPath);
+            Assert.That(obj, Is.Not.Null);
+            var go = obj as GameObject;
```
lassond: Is there a way to dig into the GameObject a little more to confirm it is coming from an Alembic file? Look for the "streaming" component for example?
Something  Alexis can take care of in a future pull request

lassond: Idea: for regular exports (not one-shot) we could also test for the number of frames through the max value for the time property

vkovec: good ideas!

PR #152, opened 2018-07-10
Uni 52375 rebuild windows dll update build steps

vkovec:


Plugin/README.txt
```diff
@@ -22,7 +22,69 @@ NOTE: source tree requires submodules
...
+
+1. follow instructions (Build Instructions 1-7) at https://portal.hdfgroup.org/display/support/Building+HDF5+with+CMake
+
+Note: make sure the binary is built in Release mode and is 64-bit
+
+2. This will generate a zip file. Unzip the file. The binaries will be located under lib/ and header files under include/
+
+3. copy the include/ directory into Plugin/external/HDF5
+
+4. copy the following library files to Plugin/external/HDF5/lib64: libhdf5.lib, libhdf5_hl.lib, libszip.lib
```
benoithudson: Why install into an HDF5 directory instead of just into Plugin/external/lib64 and Plugin/external/bin and so on?

Plugin/README.txt
```diff
@@ -22,7 +22,69 @@ NOTE: source tree requires submodules
...
+Note: should not need to build OpenEXR (steps 3-5 in README.cmake.txt)
+
+3. copy *.lib files from ${builddir}/lib to Plugin/external/ilmbase/lib64
+4. rename *.lib files (except for Half.lib) to *-2_2.lib
+5. copy ${builddir}/include directory to Plugin/external/ilmbase/
+
+Building Alembic:
+
+1. download and build zlib (v1.2.11): http://www.zlib.net/ (use cmake to build, no special args needed)
+2. after building zlib, copy zlibstatic.lib from {builddir}/Release/ to Plugin/external/HDF5/lib64
```
benoithudson: HDF5 includes zlib I think; I wonder if we can just use the hdf5 library instead?

If not, then we should build HDF5 after zlib, and tell HDF5 about our custom zlib.

Plugin/README.txt
vkovec: it was how it was setup in the zip file I was using before. Thought I would leave the same layout. Otherwise I think I would need to modify the visual studio project as well.

Plugin/README.txt
vkovec: HDF5 doesn't build zlibstatic.lib

Plugin/README.txt
vkovec: should I change it so that they should all be placed into Plugin/external/lib64 and Plugin/external/include?

PR #156, opened 2018-07-17
Added review to the docs

moujaune:
Added review to the docs

### DIFF
--- a/AlembicImporter/Packages/com.unity.formats.alembic/Documentation~/com.unity.formats.alembic.md
+++ b/AlembicImporter/Packages/com.unity.formats.alembic/Documentation~/com.unity.formats.alembic.md
@@ -2,6 +2,8 @@
 
 Use the Alembic package to import [Alembic](http://www.alembic.io/) files into your Unity Scenes. The Alembic format bakes animation data into a file so that you can stream the animation in Unity directly from the file. This saves a tremendous amount of resources, as the modeling and animating does not happen directly inside Unity.
 
+> Consider briefly mentioning that you can also export as Alembic in the first few lines since it's never mentioned until you get to that section.
+
 The Alembic package brings in vertex cache data from a 3D modeling software, such as facial animation (skinning) and cloth simulation (dynamics). When you play it back inside Unity, it looks exactly the same as it did in the 3D modeling software. 
 
 The Alembic package supports importing and playing back Meshes, Points and Cameras.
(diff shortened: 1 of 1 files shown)


AlembicImporter/Packages/com.unity.formats.alembic/Documentation~/com.unity.formats.alembic.md
```diff
@@ -103,7 +107,7 @@ The Timeline can be used to playback and record Alembic animation including:
 * Create sequences using Alembic clips with the ability to trim times and adjust clip-ins
 * Record Alembic data directly from the Timeline to an Alembic file
 
-
+> Consider making sure that you can actually edit Alembic Shot Clips via code or otherwise add it to known limitations
```
vkovec: where does it say  you can edit Alembic Shot Clips via code?

AlembicImporter/Packages/com.unity.formats.alembic/Documentation~/com.unity.formats.alembic.md
```diff
@@ -78,7 +82,7 @@ To configure a scene to export an Alembic file, add the [Alembic Exporter compon
 The component can be configured to export the entire scene or individual object hierarchies.
 
 Using the AlembicExporter component automatically disables Draw Call Batching. If the Mesh group is valid after being batched then it will be exported. In some cases the data will be batched multiplied times and the results may change.  If you want to control the Batch settings they can be found in the Rendering section of Player Settings.
-
+> Consider clarifying section; maybe a reminder line about what batching does for people who mark everything as static in their scene for performance and panic when the exporter turns it off and the project starts to slow down
```
vkovec: please give an example of what you would add

moujaune: "Adding an AlembicExporter component to your scene will disable draw call batching in your project. With batching enabled, your scene's geometry merges itself in various configurations every frame, thus somewhat marring the faithfulness of the exporter's geometry recording features.

After you're done recording, you can re-enable both static and dynamic batching for your project in the Rendering section of the Player Settings. Having the exporter run with batching enabled will engender undesirable results such at batched meshes being recorded multiple times per frame and overlapping in your Alembic file in an unsightly fashion."

AlembicImporter/Packages/com.unity.formats.alembic/Documentation~/com.unity.formats.alembic.md
vkovec: ah ok. We already mention that there is no external API though. Would that not cover this as well?

AlembicImporter/Packages/com.unity.formats.alembic/Documentation~/com.unity.formats.alembic.md
vkovec: thanks!

PR #155, opened 2018-07-16
merge stable into master

vkovec:
- update master to be the same as stable

PR #160, opened 2018-07-20
Reorg/build refactor for Katana

mjmvisser:
- Re-org'd to expected package layout for Katana
- Refactored CMakeLists to build on Windows
- Moved OpenExr and Alembic dependencies to submodules
- Committed binaries for HDF5 dependencies
- Moved  dependencies to External
- Added build.py/build.sh/build.cmd for Katana automation tools
- Updated gitlab-ci.yml to conform to Katana automation tools expectations

I kept the reorg separate from modifications, but github doesn't seem to recognize the renames in the diff below. :-(


i-saint:
there is a similar PR ( #155 ). can you merge both?

vkovec:
looks good! Is it possible to build locally with build.cmd?

mjmvisser:
> there is a similar PR ( #155 ). can you merge both?
@i-saint I don't understand. PR #155 targets master - this one targets stable. After this one lands, #155 will need to be rebased. 

> Is it possible to build locally with build.cmd?
@vkovec yes, that's how I've been testing. Could you pull this branch and check that the build works for you? I get an error on our build slaves about not finding HDF5_LIBRARIES, and I'd like to know if you get the same error. thx

PR #181, opened 2018-08-06
Uni 55220 run export tests in standalone

vkovec:
I left the tests so that they can be run in both the editor or standalone, however with `-batchmode` they will only work in standalone. In order to force standalone, the following can be added either to the test class or to individual tests:

```
[UnityPlatform(exclude = new[] { RuntimePlatform.WindowsEditor, RuntimePlatform.LinuxEditor, RuntimePlatform.OSXEditor })]
```

Here are the command line arguments used for running the tests:
/path/to/Unity.exe -batchmode -projectPath path/to/repo/TestProjects/AlembicImporter/ -forgetProjectPath -automated -testResults path/to/repo/TestResults_playmode_AlembicImporter.xml -logFile path/to/repo/editor_playmode_AlembicImporter.log -runTests **-testPlatform StandaloneWindows64**

Note: for the test platform I had to use StandaloneWindows64, the other options are at the bottom of this page: https://docs.unity3d.com/Manual/PlaymodeTestFramework.html

### DIFF
--- a/TestProjects/AlembicImporter/Assets/Tests/Runtime/UnitTests/AlembicExportTest.cs
+++ b/TestProjects/AlembicImporter/Assets/Tests/Runtime/UnitTests/AlembicExportTest.cs
@@ -118,9 +118,9 @@ namespace UnityEditor.Formats.Alembic.Exporter.UnitTests {
 #endif
         }
 
+#if UNITY_EDITOR
         // Helper for the tear-down. This is run from the editor's update loop.
         void DeleteOnNextUpdate () {
-#if UNITY_EDITOR
             EditorApplication.update -= DeleteOnNextUpdate;
             try {
                 Directory.Delete (filePath, recursive : true);
@@ -128,8 +128,8 @@ namespace UnityEditor.Formats.Alembic.Exporter.UnitTests {
             } catch (IOException) {
                 // ignore -- something else must have deleted this.
             }
-#endif
         }
+#endif
 
     }
 
@@ -151,6 +151,9 @@ namespace UnityEditor.Formats.Alembic.Exporter.UnitTests {
         /// </summary>
         /// <param name="abcPath"></param>
         private void TestAbcImported (string abcPath) {
+            // AssetDatabase is only available in Editor,
+            // make sure it doesn't run in standalone test
+            // otherwise test will fail.
 #if UNITY_EDITOR
             AssetDatabase.Refresh ();
 #endif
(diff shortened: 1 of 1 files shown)


TestProjects/AlembicImporter/Assets/Tests/Runtime/UnitTests/AlembicExportTest.cs
```diff
@@ -109,21 +104,31 @@ public class AlembicTestBase {
...
+                Directory.Delete (filePath, recursive : true);
+            } catch (IOException) {
+                // ignore -- something else must have deleted this.
+            }
+#endif
         }
 
         // Helper for the tear-down. This is run from the editor's update loop.
         void DeleteOnNextUpdate () {
+#if UNITY_EDITOR
```
inwoods: the entire function can be within DeleteOnNextUpdate as it would be a mistake to run it in non-editor mode.

TestProjects/AlembicImporter/Assets/Tests/Runtime/UnitTests/AlembicExportTest.cs
```diff
@@ -146,25 +151,34 @@ public class AlembicExportTest : AlembicTestBase {
         /// </summary>
         /// <param name="abcPath"></param>
         private void TestAbcImported (string abcPath) {
+#if UNITY_EDITOR
```
inwoods: can you explain why we don't do this in standalone?

TestProjects/AlembicImporter/Assets/Tests/Runtime/UnitTests/AlembicExportTest.cs
```diff
@@ -146,25 +151,34 @@ public class AlembicExportTest : AlembicTestBase {
...
             AssetDatabase.Refresh ();
+#endif
 
             var absPath = GetAssetsAbsolutePath (abcPath);
 
             Assert.That (string.IsNullOrEmpty (absPath), Is.False);
             Assert.That (File.Exists (absPath));
 
+#if UNITY_EDITOR
```
inwoods: why are we skipping this test?

vkovec: the asset database is only available in the editor, can't import an alembic file in standalone

inwoods: please add comment to that effect.

TestProjects/AlembicImporter/Assets/Tests/Runtime/UnitTests/AlembicExportTest.cs
inwoods: please add comment to that effect

TestProjects/AlembicImporter/Assets/Tests/Runtime/UnitTests/AlembicExportTest.cs
inwoods: make this function only visible in editor mode so that we get a coding error in non-editor mode.

vkovec:
closed because Mark found a flag to disable batchmode and make the tests work in editor

PR #182, opened 2018-08-06
Uni 54830 min public api

alexismorin:
Five new test grouped by category + a couple of stripped down helper functions.


AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,109 @@
...
+using UnityEngine.TestTools;
+
+namespace UnityEditor.Formats.Alembic.Exporter.UnitTests {
+    public class AlembicAPITest {
+        // Loads a given scene
+        IEnumerator SceneLoader (string sceneToLoad) {
+            SceneManagement.EditorSceneManager.LoadScene (sceneToLoad, UnityEngine.SceneManagement.LoadSceneMode.Single);
+            yield return null;
+        }
+        // Sets up generic export location
```
vkovec: how come you don't use the random file path generator from the test base?

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,109 @@
...
+            recorderTarget.ProcessRecording ();
+            yield return null;
+            recorderTarget.EndRecording ();
+        }
+        // Test Capture Components
+        [UnityTest]
+        public IEnumerator PublicCaptureComponents () {
+            SceneLoader ("TestAPI");
+            using (AlembicRecorder recorder = new AlembicRecorder ()) {
+                var conf = SetUpExport (recorder);
```
vkovec: this variable doesn't seem to be used. Why does SetUpExport return it?

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,109 @@
...
+        // Test AeConfig Settings
+        [UnityTest]
+        public IEnumerator PublicAeConfig () {
+            SceneLoader ("TestAPI");
+            using (AlembicRecorder recorder = new AlembicRecorder ()) {
+                var conf = SetUpExport (recorder);
+
+                conf.FrameRate = 24;
+                conf.ScaleFactor = 1;
+                conf.SwapHandedness = true;
```
vkovec: after you modify the config I think you have to set the config on the recorder: 
```
recorder.settings.Conf = conf
```
 (not 100% about this though, so please verify if possible)

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,109 @@
...
+            yield return null;
+        }
+        // Test Archive types
+        [UnityTest]
+        public IEnumerator PublicAeArchiveType () {
+            SceneLoader ("TestAPI");
+            using (AlembicRecorder recorder = new AlembicRecorder ()) {
+                var conf = SetUpExport (recorder);
+                conf.ArchiveType = UnityEngine.Formats.Alembic.Sdk.aeArchiveType.Ogawa;
+                conf.ArchiveType = UnityEngine.Formats.Alembic.Sdk.aeArchiveType.HDF5;
```
vkovec: why set it twice? Can you set the archive type in the above test (PublicAeConfig) instead? I don't think this needs a separate test

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,109 @@
...
+                Record (recorder);
+            }
+            TestAndScrub ();
+            yield return null;
+        }
+        // Test AeConfig set and load
+        [UnityTest]
+        public IEnumerator PublicAeConfigSaveLoad () {
```
vkovec: what is this testing? not sure I understand

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,109 @@
...
+                Record (recorder);
+            }
+            TestAndScrub ();
+            yield return null;
+        }
+        // Test Dispose current recording
+        [UnityTest]
+        public IEnumerator PublicDispose () {
+            SceneLoader ("TestAPI");
+            using (AlembicRecorder recorder = new AlembicRecorder ()) {
```
vkovec: this will automatically call dispose at the end of the curly brace. Let's try to simplify this test a bit:

```
AlembicRecorder recorder;
using(recorder = new AlembicRecorder()){}
// try doing recorder.settings or something else, assert it gives an exception

recorder = new AlembicRecorder();
recorder.Dispose();
// try doing recorder.settings or something else, assert it gives an exception
```

We don't need to do the full export for this one.

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+        // Sets up generic export location
+        aeConfig SetUpExport (AlembicRecorder recorderTarget) {
+            recorderTarget.settings.OutputPath = "Assets/API_UnitTests/TestAPI.abc";
+            return recorderTarget.settings.Conf;
+        }
+        // Deletes Test File
+        void TestAndScrub () {
+            Assert.That (Application.dataPath + "/API_UnitTests/TestAPI.abc", Is.Not.Null);
+            File.Delete (Application.dataPath + "/API_UnitTests/TestAPI.abc");
```
vkovec: if you use the test base to generate the filepath then you won't need to destroy it as it will be destroyed automatically

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,109 @@
...
+        IEnumerator Record (AlembicRecorder recorderTarget) {
+            var result = recorderTarget.BeginRecording ();
+            Assert.That (result, Is.True);
+            recorderTarget.ProcessRecording ();
+            yield return null;
+            recorderTarget.EndRecording ();
+        }
+        // Test Capture Components
+        [UnityTest]
+        public IEnumerator PublicCaptureComponents () {
```
vkovec: please change the names of the tests to start with "Test" and remove "Public". e.g. this test would become "TestCaptureComponents"

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
moujaune: Just tested, awesomely enough you don't!

AlembicImporter/Assets/API_UnitTests/AlembicAPITest.cs
```diff
@@ -0,0 +1,84 @@
...
+            yield return null;
+        }
+
+        // Test Dispose current recording
+        [UnityTest]
+        public IEnumerator TestDispose () {
+            SceneLoader ("TestAPI");
+            AlembicRecorder recorder;
+            using(recorder = new AlembicRecorder()){}
+            Assert.That(recorder.settings, Is.Not.Null);
```
vkovec: does this pass the test? I thought it would throw an exception since its accessing a disposed object.

moujaune:
EditorBuildSettings.asset and ProjectSettings.asset were changed when I added the new testAPI scene to the build settings!

AlembicImporter/Assets/UnitTests/AlembicExportTest.cs
```diff
@@ -160,11 +147,28 @@ public class AlembicExportTest : AlembicTestBase {
...
+            return alembicExporter;
+        }
+
+
+
+        /// <summary>
+        /// Test that the abc file was reimported and can be loaded into Unity through
+        /// the asset database.
+        /// </summary>
+        /// <param name="abcPath"></param>
```
vkovec: the function moved but it's description stayed here :P. Please move the description too.

AlembicImporter/Assets/UnitTests/AlembicExportTest.cs
```diff
@@ -132,6 +132,12 @@ public class AlembicTestBase {
             yield return null;
         }
 
+        /// <summary>
+        /// Test that the abc file was reimported and can be loaded into Unity through
+        /// the asset database.
+        /// </summary>
+        /// <param name="abcPath"></param>
+
```
vkovec: please remove this space

proto.com.unity.formats.alembic/Runtime/Scripts/Misc/Bool.cs
```diff
@@ -3,10 +3,12 @@
 
 namespace UnityEngine.Formats.Alembic.Sdk
 {
-    // bool is marshal as int (4 byte) by default and you need ugly [MarshalAs(UnmanagedType.U1)] to pass to (or receive from) C++ code.
-    // this struct emulates bool and marshal as byte (1 byte). this makes things bit easier in some cases.
+    /// <summary>
+    /// bool is marshal as int (4 byte) by default and you need ugly [MarshalAs(UnmanagedType.U1)] to pass to (or receive from) C++ code.
```
benoithudson: `Alembic marshals bool as 1 byte, whereas the C# default is to marshal bool as 4 bytes. This struct implements 1-byte marshalling.`

proto.com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
```diff
@@ -72,22 +91,28 @@ public bool CaptureMeshRenderer
...
         [SerializeField]
         private bool captureSkinnedMeshRenderer = true;
-        public bool CaptureSkinnedMeshRenderer
+        internal bool CaptureSkinnedMeshRenderer
         {
             get { return captureSkinnedMeshRenderer; }
             set { captureSkinnedMeshRenderer = value; }
         }
-
+
```
benoithudson: oops, please remove this.

PR #189, opened 2018-08-21
Uni-56835 V1.0 - doc changes

vkovec:


inwoods:
why not prepare a git patch so that you can replay the changes on stable - then we can scrap the v1.0 branch?

vkovec:
whats the difference?

PR #212, opened 2018-11-23
don't over-write user specified time range on re-import.

jasonm-unity:
- prevent stream player inspector for allowing edits on objects that are marked not-editable.


mjmvisser:
Tested locally, lgtm.

PR #223, opened 2019-01-07
FTV-36 [Alembic] Alembic player is bloating the scene by storing meshes as prefab overrides

clusty:
The alembic importer generates a Prefab. Once this is instantiated in the scene, the AlembicStreamPlayer replaces all meshes with the current correct version at the current time. This is bloating the scene as this counts for a prefab override.
The fix is marking all the MeshFilters with DoNotSave when instantiated in the scene. This prevents Scene size bloating.


CHANGELOG.md
```diff
@@ -1,12 +1,15 @@
 # Changes in Alembic for Unity
 
+## [1.0.0-preview.9] - 2019-01-03
+### Changes
+- Meshes generated by the AlembicSreamPlayer are not serialized with the Scene
```
mjmvisser:
```suggestion
- Meshes generated by the AlembicStreamPlayer are not serialized with the Scene
```

CHANGELOG.md
```diff
@@ -1,12 +1,15 @@
...
+
 ## [1.0.0-preview.8] - 2018-12-19
 ### Changes
 - Alembic files are No longer copied to the StreamingAssets folder
 - renamed Linux dso to abci.so for consistency
 - reordered DllImport path selection so Editor locations are always found first
 
-### Known Issues
```
mjmvisser: why was this line removed?

clusty: I think this is a typo. Will fix it.

clusty:
@i-saint This code needs to be conditional and only executed when instantiating in the Scene. The importer also calls AlembicMesh.AddMeshComponents, and should not mark components as DoNotSave (if set, this breaks thumbnails and meshes are not actually imported, thus they are not "addressable" outside of the ABC). 
Would you like me to move my code inside AlembicMesh.AddMeshComponents, but propagate the firstImport flag deeper ?

i-saint:
@clusty understood. I have no question other than that. I approve.

jcowles:
Looks like the change only applies to MeshFilters, should it also apply to SkinnedMeshRenderers (which also have a sharedMesh)?

mfe:
Tested it. LGTM.

jcowles:
FYI, I tried this same logic in USD and I'm seeing strange bugs where **SOMETIMES** the scene will load and all the meshes will be destroyed. The condition is somehow related to selection.

Just a heads up.

clusty:
Thanks a bunch. Will keep an eye out.

PR #224, opened 2019-01-07
HDRP Motion Vector Pass | Alembic Velocity Support

johnpars:
Contains an addition of two files which, in a similar fashion to the existing legacy version, is a near-copy of their original (Lit in this case), with small modification of the MotionVectors pass to properly integrate the runtime-calculate alembic velocities, submitted via TEXCOORD3. 

**NOTE:**

**1.)** Adding motion vector support to an alembic asset in your scene is controlled by the linked material. Currently, the shader is found under `HDRenderPipeline` -> `LitAlembic` (though `Alembic` -> `Lit` may be more suitable).

**2.)** By default, HDRP disables the object motion vector pass on materials. Currently  `Enable MotionVector For Vertex Animation` must be selected for the MotionVector pass to be included (see image). 

It is possible to change the default value in ShaderLab, however since the property is already serialized on the creation of a new material in HDRP, the compiler will defer to that instead, leaving the pass disabled.

Inheriting a custom UI from the Lit shader to solve this is also currently not simple, as it exists outside of the HD assembly. 

image

**2.)** Adding this renderpipeline support leaves more to be desired in the organization of the Shader/ folder, perhaps organization by renderpipeline.


i-saint:
looks good to me. thank you very much!

Verasl:
Since this is a copy of the lit shader, I'm taking there is no ShaderGraph support with this PR?

johnpars:
Hey Andre, that's correct- this PR does not also contain the SG implementation. 
To add it, I believe it would take a similar copy as the Lit SG implementation, but link to the velocity pass that is included in this PR. 

There is a discussion to be had, however, if it's not simpler to have the alembic velocity integration be implicit to a render pipeline's primary velocity pass. In the end, all that is required is an extra vertex input + transformation to integrate it to the velocity calculation, and that is all this PR is doing.

sebastienlagarde:
Hey John, 

So from what I understand Alembics object provide their precomputed 3d object space velocity that you need to substract to current object space position. Right?

So first question. From what I understand from the PR, you don't have need of supporting skinning or additional vertex animation on top of the Alembic, right ?

In this case, instead of using an additional alembicVelocity, you could simply store in the UVSet4 the positionOS - alembicVecloity and use the PreviousVelocity slot ?
then the goal is to enable hasDeformation with unity_MotionVectorsParams.x > 0.0; This is where you may require a C++ change. But doing this allow you to reuse the original lit shader without modification.
Otherwise we could add an option to force hasDeformationto be true, easy to do.

If however you want to support additional skinning and vertex animation, then we need both previousPositionOS and AlembicVelocity. In this case we could write:

float3 effectivePositionOS = (hasDeformation ? inputPass.previousPositionOS : inputMesh.positionOS)
effectivePositionOS -= inputPass.alembicVelocity; // <= this line

#if defined(HAVE_MESH_MODIFICATION)
        AttributesMesh previousMesh = inputMesh;
        previousMesh.positionOS = effectivePositionOS ;

        previousMesh = ApplyMeshModification(previousMesh);
        float3 previousPositionRWS = TransformPreviousObjectToWorld(previousMesh.positionOS);
#else
        float3 previousPositionRWS = TransformPreviousObjectToWorld(effectivePositionOS);
#endif

In this case I could add a new shader features (#define ADDITONAL_VECLOCITY_CHANGE and use a generic name like "additionalVeclocityChange") to add it to the original lit.shader and guard the single extra line, so you don't need an extra Alembic shader but can simply check a box.

Tell me what you think

Side note: A disabled motion vector pass doesn't mean there is no motion vector pass (due to how legacy unity have implemented motion vector). Please be careful with that. See documentation here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/wiki/How-motion-vector-work

johnpars:
Hi Sebastien,

Thank you for your notes- now, on to your questions:

> So from what I understand Alembics object provide their precomputed 3d object space velocity that you need to substract to current object space position. Right?

That's correct- the Alembic velocities are currently computed on the alembic runtime and sent down to GPU.

> So first question. From what I understand from the PR, you don't have need of supporting skinning or additional vertex animation on top of the Alembic, right ?

I will leave the final answer of this to @mjmvisser- the direction we go is likely based on this. As this current PR stands, indeed there is not support for the additional mesh modification / skinning.

I think both options have their pros and cons. Being able to nicely integrate the alembic velocities directly with previous position in UVSet4 is a clean idea, that also would make it a simpler process to support alembic motion vectors in other render pipelines. However for thoroughness, the second option holds just as much weight. Will have to see what Mark thinks before knowing for sure. 

> Note that the change I propose is for all material (stacklit, Axf, hair, fabric, lit - including shader graph version), not only for Lit.

Understood, this is the ideal situation I believe.

> But I expect you want to try first your PR see if everything works as you want before taking any choice?

We shipped the ADAM 2 and 3 productions using the same idea in this PR, and have made sufficient testing in HD. Would it also be a good opportunity to implement a graphics test? Any links to this process would be great (I recall there being a testing standard for HD features / proposals)

sebastienlagarde:
Ok, poke me when a decision is taken (for support of additional skinning), and we can integrate the change the best way.

sebastienlagarde:
>Any links to this process would be great 
A bit old, but still useful: https://docs.google.com/document/d/1n4d0BI_8BdLwnH-OMUAAAS8z2RAwJ32i-6MgPwFJDgs/edit?usp=sharing

PR #226, opened 2019-01-15
Timeline fixes

clusty:
This PR implements 2 minor fixes with Timeline:
*Time changes from timeline AlembicShotClips are not persistent in the scene anymore
*AlembicRecorderClip is using now Tags to target GameObject roots for recording (before it was storing GO paths, which did not work very well with multiScene workflows). There is a Warning in the clip inspector if there are multiple GO that match the tag.


jcowles:
I see that the paths used to be strings -- do [ExposedReference<>](https://docs.unity3d.com/ScriptReference/ExposedReference_1.html) objects not work well with the multi-scene workflow either? I guess it would make sense if they did not.

clusty:
They do not, unfortunately. ExposedRefs are much more elegant compared to the Tag based workflow...

PR #233, opened 2019-03-20
Remove linux32

clusty:
This PR removes the usage of BuildTarget.StandaloneLinux for 2019.2+ since it's being obsoleted.
Small side cleanup, removed all linux mentions form asmdef-s (should have been done in the initial commit since it's not compatible with linux today.)


clusty:
Closing to restart one with CI enabled.

PR #238, opened 2019-03-29
Windows file locks ftv-142

clusty:
This PR changes the windows implementation for file access.
On windows there are certain apps (P4) that are unable to update the ABC assets when they are opened in Unity because of the OS locks. This implementation uses windows APIs to open files without locks.

### DIFF
--- a/Source/abci/Importer/aiContext.cpp
+++ b/Source/abci/Importer/aiContext.cpp
@@ -4,6 +4,11 @@
 #include "aiObject.h"
 #include "aiAsync.h"
 #include <istream>
+#ifdef WIN32
+    #include <windows.h>
+    #include <io.h>
+    #include <fcntl.h>
+#endif
 
 static std::wstring L(const std::string& s)
 {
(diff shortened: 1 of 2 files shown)


Source/abci/Importer/aiContext.cpp
```diff
@@ -39,7 +39,28 @@ static std::string NormalizePath(const char *in_path)
...
 
+#ifdef WIN32
 
+class lockFreeIStream : public std::ifstream
+{
+private:
+	FILE *_file = nullptr;
+public: 
+	lockFreeIStream(const wchar_t  *name) :_file( _wfsopen(name, L"rb", _SH_DENYNO)), 
+                                               std::ifstream(_file)
```
jeanblouin: You are opening the file with _SH_DENYNO, Which means as we read the file, someone can also read and write to it. What would happen if someone write to this file while you are reading from it?

Source/abci/Importer/aiContext.cpp
```diff
@@ -39,7 +39,28 @@ static std::string NormalizePath(const char *in_path)
...
 }
 
+#ifdef WIN32
 
+class lockFreeIStream : public std::ifstream
+{
+private:
+	FILE *_file = nullptr;
+public: 
+	lockFreeIStream(const wchar_t  *name) :_file( _wfsopen(name, L"rb", _SH_DENYNO)),
```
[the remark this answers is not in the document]
clusty: The usage of this class is guarded with try/catch

Source/abci/Importer/aiContext.cpp
clusty: the ABC importer is detecting file delete/moves and closes all files in this case.
Need to test if I can replace while actually playing.

PR #239, opened 2019-03-29
Standalone player FTV-92

clusty:
This branch allows users to create standalone players with Alembic that are also redistributable.
To allow this, the alembics are copied into the streaming assets of the player build and all paths are redirected to them.
To eliminate potential name clashes, these files are copied under a hashed name. (hash is dependent on path/name in the project)

### DIFF
--- a/com.unity.formats.alembic/Editor/Importer/AlembicBuildPostProcess.cs
+++ b/com.unity.formats.alembic/Editor/Importer/AlembicBuildPostProcess.cs
@@ -1,0 +1,101 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using UnityEditor.Build;
+using UnityEditor.Build.Reporting;
+using UnityEditor.Callbacks;
+using UnityEditor.SceneManagement;
+using UnityEngine;
+using UnityEngine.Formats.Alembic.Importer;
+using UnityEngine.SceneManagement;
+using Object = UnityEngine.Object;
+
+namespace UnityEditor.Formats.Alembic.Importer
+{
+    public static class AlembicBuildPostProcess
+    {
+        internal static readonly List<KeyValuePair<string,string>> FilesToCopy = new List<KeyValuePair<string,string>>();
+        [PostProcessBuild]
+        public static void OnPostProcessBuild(BuildTarget target, string pathToBuiltProject)
+        {
+            if (target != BuildTarget.StandaloneOSX && target != BuildTarget.StandaloneWindows64)
+            {
+                return;
+            }
+
+            foreach (var files in FilesToCopy)
+            {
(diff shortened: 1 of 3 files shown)


com.unity.formats.alembic/Editor/Importer/AlembicBuildPostProcess.cs
```diff
@@ -0,0 +1,109 @@
...
+
+namespace UnityEditor.Formats.Alembic.Importer
+{
+    public static class AlembicBuildPostProcess
+    {
+        internal static readonly List<KeyValuePair<string,string>> FilesToCopy = new List<KeyValuePair<string,string>>();
+        [PostProcessBuild]
+        public static void OnPostProcessBuild(BuildTarget target, string pathToBuiltProject)
+        {
+            if (target != BuildTarget.StandaloneOSX && target != BuildTarget.StandaloneWindows64)
```
mfe: you're doing this condition twice, how about factorizing it in a function so if/when we add Linux Standalone there will be less changes to do ?

com.unity.formats.alembic/Editor/Importer/AlembicBuildPostProcess.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        static string GetStreamingAssetsPath(BuildSummary summary)
+        {
+            switch (summary.platform)
+            {
+                case BuildTarget.StandaloneOSX:
+                    return Path.Combine(summary.outputPath, "Contents/Resources/Data/StreamingAssets");
+                case BuildTarget.StandaloneWindows64:
+                    var name = Path.ChangeExtension(summary.outputPath, null);
+                    return name+"_Data/StreamingAssets";
```
mfe: Some formatting issue in this file. Here, add spaces around '+'.
Also remove lines 36, 55 and 90 (empty lines with tabs).

com.unity.formats.alembic/Editor/Importer/AlembicBuildPostProcess.cs
```diff
@@ -0,0 +1,109 @@
...
+            
+            FilesToCopy.Clear();
+        }
+    }
+
+    class AlembicProcessScene : IProcessSceneWithReport
+    {
+        public int callbackOrder
+        {
+            get { return 9999;}
```
mfe: Is that the usual way to be sure it's the last ?

clusty: No Other way I could find. 
The order is marginally important: if users were to instantiate more Alembics after I fixed the ones in scenes, they would simply not work in standalone builds....

PR #254, opened 2019-04-15
FTV-132 : Recording objects that get destroyed at runtime stops recor…

clusty:
…ding

### DIFF
--- a/com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
+++ b/com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
@@ -417,9 +417,13 @@ namespace UnityEngine.Formats.Alembic.Util
 
             public override void Setup(Component c)
             {
-                var target = c as Transform;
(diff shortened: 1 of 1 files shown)


com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
```diff
@@ -417,9 +417,13 @@ class TransformCapturer : ComponentCapturer
 
             public override void Setup(Component c)
             {
-                var target = c as Transform;
-                abcObject = parent.abcObject.NewXform(target.name + " (" + target.GetInstanceID().ToString("X8") + ")", timeSamplingIndex);
-                m_target = target;
+                if (parent != null && c is Transform)
```
mfe: When parent can be null ? and when can setup of a TransformCapturer be called with something that is not a Transform ?
Could this be also an issue with all Capturers ?

clusty: The parent can be null because it was deleted. 
The parent cannot be different than a transform really. I am just checking my assumptions. Previous code was casting and never checking the result.

clusty:
Will make a different fix. There are more problems with the handling of changing parenting structure.

PR #259, opened 2019-04-23
FTV-83 Physical Cam import export

clusty:
This branch uses the physical parametrisation of Cameras on import and export.
The code is also factored out to use the same data structures for import and export.
The I/O scale impacts the near far parameters. The sensor size in alembic is defined in cm, thus the 10/0.1 factors for import and export.


com.unity.formats.alembic/Runtime/Scripts/Misc/CameraData.cs
```diff
@@ -0,0 +1,14 @@
+using System.Runtime.InteropServices;
+
+namespace UnityEngine.Formats.Alembic.Sdk
+{
+    [StructLayout(LayoutKind.Sequential)]
+    struct CameraData
+    {
+        public Bool visibility { get; set; }
+        public float focalLength { get; set; }
+        public Vector2 sensorSize { get; set; }
```
mfe: Not very fond of converting those in Vectors. It confuses the story.
And you end up with stuff that does not make sense : `m_camera.nearClipPlane = m_abcData.nearFar.x`, near is not an X, neither far is a Y. And having `nearFar[0]` is not very graceful either.
Code is written once but then read a million time by various developers.

So I would either keep the separated variables or have some struct or named tuples that would allow planClipping.near, planClipping.far, lensShift.horizontal...

Was there a specific reason to use vectors ?

clusty: I usually pack things that go together as the same structure. If you edit a near, it's highly likely you'll edit the far. 
Also makes marshalling code easier to write since I don't have the problem of variable ordering.
Tuples are doing "type pollution" which is overkill in this context (not to mention they are a C# 7 feature that is still not widely supported)

Issue #210, opened 2018-11-21
Vertex Color is missing - alembic export from Houdini 16-17


gasanoff:
note: it might be the doubling of #86 or similar issues, but I couldn't find the solution yet.

- static .abc file  https://drive.google.com/open?id=1N5aNH9GIG-1xIXC61ERbeifU4lmUYXF3
- AlembicForUnity ver. 20180413
- Editor Log + additional screenshots with the issue
https://drive.google.com/open?id=17G6LMo5L-O51Rkz41pXP1SPNJKrm93a2

__Issue: ABC exported from Houdini 16-17 has vertex color information but it's not properly named/stored to be seen in Unity__

Amplify Shader or Unity Shader Graph have the same result when applying a Vertex Color Node - white/gray solid color. But it works fine in with FBX geometry exported from 3ds max though (https://drive.google.com/open?id=1zIZxaJNQk7RTFy0N5VHPhw6kqygJ_twJ)

ABC's vertex colors can be rendered in 3dsMax by selecting a proper channel # for vertex colors
https://drive.google.com/open?id=1KuvBJOxKqDO19xXKItwvWvEF7bZErjDO

_____________
Here's the VertexColors from Houdini to Unity at SideFX forums
https://www.sidefx.com/forum/topic/39048/

Still waiting for any comments about the proper exporting/naming all the datasets/groups/channels from Houdini team

Thank you!

i-saint:
thanks for the valuable report.

I investigated your abc file and confirmed Houdini's vertex color doesn't work with current version of AlembicForUnity. supporting it requires some modifications in plugin (C++ side).

unfortunately, vertex color is not a part of Alembic's specification. DCC tools support it as an extension attribute in their own convention. current version of AlembicForUnity supports at least Maya and 3ds Max's vertex color attribute, but Houdini's one is very different structure and can not be recognized. I will try to support it...

gasanoff:
Seiya, thank you so much!

I'm still trying to get any help from SideFX - there might be a way of proper exporting to Alembic from Houdini 17. 
Ready to test/export data with multiple settings anytime

cguer:
Fixed in version 1.0.6

sambeebe:
Hey, I see this was fixed but am having a lot of trouble getting vertex colors to display in Unity when being exported from Houdini. What is the proper method of doing so, and is a custom shader necessary? Thanks!

PR #271, opened 2019-05-23
Houdini vertex colors - GH #210

cpredmann:
Houdini stores Cd as a C3f of rgb, with alpha as a separate field, this addresses it by allowing for that value to be used as Mesh color in Unity.
* Adds an rgb member that will hold an IC3f arbGeomAttrib found on the imported mesh
* If no IC4f is found on the mesh, it will set the Unity Mesh.Colors to the rgb values (with 1.0 for alpha)
* Rename colors to rgba, for clarity


clusty:
Closing and moving to: https://github.com/unity3d-jp/AlembicForUnity/pull/280

PR #276, opened 2019-08-08
Ftv 145 shared normals

clusty:
This is the same PR as Chris' but with a cleaner history and some dead code cleanup.

* Calculate normals on original topology
* Remap normals to refined mesh
* Linux name fix

### DIFF
--- a/Source/abci/Importer/aiObject.cpp
+++ b/Source/abci/Importer/aiObject.cpp
@@ -58,8 +58,13 @@ aiObject::aiObject(aiContext *ctx, aiObject *parent, const abcObject &abc)
     , m_abc(abc)
     , m_parent(parent)
 {
+#if defined(__linux__)
+    m_name = m_abc.getName();
+    m_fullname = m_abc.getFullName();
+#else
     m_name = SanitizeNodeName(m_abc.getName());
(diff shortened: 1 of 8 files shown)


Source/abci/Importer/aiObject.cpp
```diff
@@ -58,8 +58,13 @@ aiObject::aiObject(aiContext *ctx, aiObject *parent, const abcObject &abc)
     , m_abc(abc)
     , m_parent(parent)
 {
+#if defined(__linux__)
```
mfe: Would have been nice to have this in a separate commit

Source/abci/pch.h
```diff
@@ -57,6 +57,7 @@ using namespace Alembic;
 
 using abcV2 = Imath::V2f;
 using abcV3 = Imath::V3f;
+using abcV3i = Imath::V3i;
```
mfe: Not sure where this is used ?

clusty: Doubt it's needed.

Source/abci/Importer/aiObject.cpp
clusty: Done

PR #278, opened 2019-08-08
Ftv 167 GameObject rename

clusty:
This is another bug related to the way prefab references are being stored.
The root gameObject name needs to be file independent. This fix sets the name once and reuses it every time.

### DIFF
--- a/com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
+++ b/com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
@@ -54,6 +54,8 @@ namespace UnityEditor.Formats.Alembic.Importer
     {
         [SerializeField]
         private string rootGameObjectId;
+        [SerializeField]
+        private string rootGameObjectName;
 
         [SerializeField]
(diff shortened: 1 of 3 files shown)


com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
```diff
@@ -49,12 +69,15 @@ public static AssetMoveResult OnWillMoveAsset(string from, string to)
         }
     }
 
-    [ScriptedImporter(5, "abc")]
+    [ScriptedImporter(6, "abc")]
     internal class AlembicImporter : ScriptedImporter
     {
         [SerializeField]
+#pragma warning disable 0649
```
clusty: This variable is always written by serialization, but the warning would cause packman clean console tests to fail.

bastienunity: For future references: https://unity.slack.com/archives/C06TPSM32/p1562336486321800

PR #287, opened 2019-09-24
Ftv 275 linux build

clusty:
This branch includes precanned linux binaries.
Summary of changes:
*CMake change to mitigate HDF5 errors on Ubuntu.
*Standalone build fix: initially disabled Alembics were not being copied into the app.
*Precanned linux binary and instructions how to build on CentOS


.gitignore
```diff
@@ -51,7 +51,7 @@ install/**
 External/install/**
 com.unity.formats.alembic/.npmignore
 com.unity.formats.alembic/package.json
-com.unity.formats.alembic/Runtime/Plugins
+#com.unity.formats.alembic/Runtime/Plugins
```
mfe: Just remove the line

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicMesh.cs
```diff
@@ -327,12 +327,12 @@ public override void AbcSyncDataEnd()
                         split.mesh.SetUVs(3, split.velocities.List);
 #endif
                     }
-
```
mfe: Does not feel that this two changes are needed. I would remove them.

.gitignore
clusty: This is temporary. As soon as I have a valid linux CI, this will be put back.

clusty:
> Minor remarks.
> 
> Can you explain why the CI is not working in the branch ? Will dev be broken too after the merge ?

The CI is failing because of a new validation suite test was added to ensure SemVer compliance: **Thou shalt not add a new supported platform during a patch release** : We are doing exactly that. I think the dev CI will be broken till this package is published: the test looks at the API delta between the last published version and version to be tested. Will confirm.

PR #290, opened 2019-10-02
Linux hdf5

clusty:
Disable the HDF5 export support on Linux temporarily until we decide if we drop support altogether.
Importing of HDF5 files will most likely fail with a generic error: "Cannot load file"
I cannot detect the file format beforehand in order to print a more specific error message.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -15,7 +15,7 @@ if(WIN32)
 elseif(APPLE)
     set(HDF5_LIBRARIES "${HDF5_LIBRARIES};-ldl;-lpthread;${HDF5_ROOT}/lib/libz.a;${HDF5_ROOT}/lib/libszip.a")
 else()
-    set(HDF5_LIBRARIES "${HDF5_LIBRARIES};-ldl;-lpthread")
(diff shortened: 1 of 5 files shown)


Source/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ if(WIN32)
 elseif(APPLE)
     set(HDF5_LIBRARIES "${HDF5_LIBRARIES};-ldl;-lpthread;${HDF5_ROOT}/lib/libz.a;${HDF5_ROOT}/lib/libszip.a")
 else()
-    set(HDF5_LIBRARIES "${HDF5_LIBRARIES};-ldl;-lpthread")
+    set(HDF5_LIBRARIES "${HDF5_LIBRARIES};-ldl;-lpthread;${HDF5_ROOT}/lib/libz.a;${HDF5_ROOT}/lib/libszip.a")
```
mfe: Do we still need this ?

clusty: > Please link this with a (documented) Jira issue ID

I created an issue: FTV-326. Will rename the branch after it's approved.

clusty: > Do we still need this ?
This is still needed because without it, if ever we try reading an HDF5 file, you'd get a crash (HDF5 lib will be complaining about dependencies). This way the file load will fail.

The whole CMAKE linux  bit can be removed in case we decide to rip the HDF5 support from linux completely.

PR #294, opened 2019-10-29
Alembic jobification

clusty:
This branch jobifies 2 parts of the Alembic update code:
* During the Update, an UpdateSamples job is launched. This reads the ABC file in native memory (1 per alembic file)
* During LateUpdate every Mesh launches a FillVertexBufferJob that converts the data from native to managed memory.

Removed the Async option as this was not gaining a lot of performance as is superseded by the UpdateSamples job.

Quick performance test:
100 instances of backflip ABC file:
before: 15FPS
after 90FPS

### DIFF
--- a/Source/abci/Importer/AlembicImporter.cpp
+++ b/Source/abci/Importer/AlembicImporter.cpp
@@ -156,9 +156,7 @@ abciAPI aiSample* aiSchemaGetSample(aiSchema * schema)
 
 abciAPI void aiSchemaUpdateSample(aiSchema* schema, const abcSampleSelector *ss)
 {
-    if (schema)
-    {
-        schema->markForceSync();
+    if (schema) {
         schema->markForceUpdate();
         schema->updateSample(*ss);
     }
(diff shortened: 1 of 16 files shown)


Source/abci/Importer/AlembicImporter.cpp
```diff
@@ -156,9 +156,7 @@ abciAPI aiSample* aiSchemaGetSample(aiSchema * schema)
 
 abciAPI void aiSchemaUpdateSample(aiSchema* schema, const abcSampleSelector *ss)
 {
-    if (schema)
-    {
-        schema->markForceSync();
+    if (schema) {
```
lucillecaillaud: Minor, curly bracket should be on a new line, like before.

Source/abci/Importer/aiSchema.h
```diff
@@ -198,10 +174,7 @@ class aiTSchema : public aiSchema
                 sample = nullptr;
         }
 
-        if (sample)
-        {
-            if (m_force_sync)
-                sample->markForceSync();
+        if (sample) {
```
lucillecaillaud: Minor, same as above, curly bracket should be on a new line, like before.

PR #295, opened 2019-10-29
FTV-257 :Submeshes are created in the order of "discovered" face-sets

clusty:
This branch creates the sub meshes in the order in which it "discovers" facesets:
This is important because we have workflows that clone the material assignments from FBX files and the meshes need to appear in the same order. This order is not give by any spec I found so far.

From the alembic SDK we receive an int for every face corresponding to the faceID:
Assume the following data for faces and faceIDs:
(f1,f2,f3,f4)->(2,2,1,1)
before:
sub mesh 1->(f3,f4)
sub mesh 2->(f1,f2)
after:
sub mesh 1->(f1,f2)
sub mesh 2->(f3,f4)

This was tested to do the right thing with multiple files exported from Maya and 3dsMax.

### DIFF
--- a/Source/abci/Foundation/aiMeshOps.cpp
+++ b/Source/abci/Foundation/aiMeshOps.cpp
@@ -1,5 +1,6 @@
 #include "pch.h"
 #include "aiMeshOps.h"
+#include <unordered_set>
 
 
 namespace impl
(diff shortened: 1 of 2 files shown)


Source/abci/Foundation/aiMeshOps.cpp
```diff
@@ -254,9 +258,13 @@ void MeshRefiner::genSubmeshes(IArray<int> material_ids)
                     {
                         int id = (int)tmp_submeshes.size();
                         tmp_submeshes.push_back({});
-                        tmp_submeshes.back().material_id = id - 1;
                     }
                     tmp_submeshes[mid].index_count += (counts[fi] - 2) * 3;
+                    if (std::find(materialSet.begin(), materialSet.end(), mid) == materialSet.end() )
```
jeanblouin: Since this is a linear search in the unordered_set perf could be problematic.

PR #297, opened 2019-11-19
API docs review

sebastienduverne:


com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
```diff
@@ -99,7 +99,7 @@ public bool AssumeNonSkinnedMeshesAreConstant
         [SerializeField]
         bool captureMeshRenderer = true;
         /// <summary>
-        /// Enable to capture Meshes.
+        /// Enable to capture Mesh components.
```
clusty: Meshes are not technically not components...

PR #298, opened 2019-11-21
FTV-332: Add Warning in the AlembicImporter with HDF5

clusty:
This PR adds a warning in the Importer inspector when a file is detected to be HDF5. 
The same warning gets written to the console on the first import.

The Alembic exporter displays the HDF5 Archive type as Obsolete.

### DIFF
--- a/Source/abci/Importer/aiContext.cpp
+++ b/Source/abci/Importer/aiContext.cpp
@@ -179,7 +179,8 @@ aiContextManager::~aiContextManager()
 }
 
 aiContext::aiContext(int uid)
-    : m_uid(uid)
+    : m_uid(uid),
+      m_isHDF5(false)
 {
 }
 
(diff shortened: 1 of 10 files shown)


Source/abci/Importer/aiContext.cpp
```diff
@@ -329,6 +330,7 @@ bool aiContext::load(const char *in_path)
             Alembic::AbcCoreOgawa::ReadArchive archive_reader(m_streams);
```
judubu: Do you know you can use the IFactory class to do the whole hdf5/ogawa dance for you?
```
  IFactory	factory;
  IFactory::CoreType	archive_type;
  m_archive = factory.getArchive(m_path.c_str(), archive_type);
  m_isHDF5 = (archive_type == IFactory::kHDF5);
```

PR #304, opened 2019-12-16
FTV-369: Export ever growing geometry

clusty:
The alembic exporter was not clearing the Index buffer and the face buffer(a buffer describing the polygon count for the current face.) at the beginning of the frame resulting in ever growing geometries.

### DIFF
--- a/Source/abci/Exporter/aePolyMesh.cpp
+++ b/Source/abci/Exporter/aePolyMesh.cpp
@@ -105,14 +105,11 @@ void aePolyMesh::writeSampleBody()
         ApplyScale(m_buf_velocities.data(), (int)m_buf_velocities.size(), scale);
     }
 
-    // if face counts are empty, assume all faces are triangles
-    if (m_buf_faces.empty())
-    {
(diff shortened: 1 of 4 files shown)


Source/abci/Exporter/aePolyMesh.cpp
```diff
@@ -105,14 +105,11 @@ void aePolyMesh::writeSampleBody()
         ApplyScale(m_buf_velocities.data(), (int)m_buf_velocities.size(), scale);
     }
 
-    // if face counts are empty, assume all faces are triangles
-    if (m_buf_faces.empty())
-    {
-        m_buf_faces.resize((int)(m_buf_indices.size() / 3), 3);
-    }
-
```
lucillecaillaud: I'm curious why you removed that and if there is a link with the PR or if it's just a parallel clean?

clusty: The array was always initialized a bit lower, so code served no obvious purpose.

PR #312, opened 2020-04-20
FTV-582: Motion vector bug

clusty:
The branch flips the direction of the motion vectors (multiply with -1): the HDRP expects vectors to point in the direction of the movement, and the ABC are backwards.
The flip is performed in a burst enabled job in C# (new dependency on Burst. was added).
PR also adds a test project that is HDRP ready.

Only proper change is:
@judubu com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicMesh.cs
@cguer pls test TAA as well motion blur.


com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicMesh.cs
```diff
@@ -302,6 +319,20 @@ public override void AbcSyncDataEnd()
...
+            {
+                var split = m_splits[i];
+                if (split.active &&  split.velocities.Count > 0)
+                {
+                    var job = new MultiplyByConstant
+                    {
+                        data = split.velocities.Pointer,
+                        scalar = -1
+                    };
+                    m_PostProcessJobs[i] = job.Schedule(split.velocities.Count, 2048);
```
judubu, line 332: Any specific reason you are using a 2048 batch size?

clusty: empirical constant tweaking.

PR #313, opened 2020-04-22
FTV-514: Crash when Activation clip starts after Alembic clip

clusty:
The class not being initialized properly.


clusty:
Does not work. Closing.

PR #315, opened 2020-04-24
FTV-514: Crash with activation

clusty:
context was being changed before the jobs that depended on it were finished.


com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
```diff
@@ -62,26 +116,25 @@ public static void ReconnectStreamsWithPath(string path)
         AlembicStreamDescriptor m_streamDesc;
         AlembicTreeNode m_abcTreeRoot;
         aiConfig m_config;
-        aiContext m_context;
+        ContextPair context;
```
judubu, line 119: Why the name change? m_context to context

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
```diff
@@ -9,15 +9,69 @@
...
 {
-    internal sealed class AlembicStream : IDisposable
+    sealed class AlembicStream : IDisposable
     {
-        public struct UpdateSamplesJob : IJob
+        /// <summary>
+        /// This class hides the context. The reason for that it that there are Jobs depending on it and we need
+        /// to ensure they get completed before any set happens on the context
+        /// </summary>
+        public struct ContextPair
```
judubu, line 18: Can I suggest SecureContext or SyncedContext? I think it conveys more meaning than ContextPair

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
clusty: I hate variables starting with m :) Fixed for consistency.

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
clusty: changed to SafeContext.

PR #318, opened 2020-04-30
FTV-394: Alembic recorder Integration

clusty:
This PR implements the Alembic integration for Unity recorder. This is mostly  a repackage of the existing component Alembic Exporter Mono Behaviour.
Screen Shot 2020-04-30 at 1 05 25 PM

The main differences  are that the Unity Recorder being an asset serializes differently the GameObject scene reference. Also the file name is managed by the recorder.
There is also a new test project that tests this integration.
Minimum recorder version: 2.2.0-preview.4


clusty:
> I don't see here any content that I can review, except (from the screenshot) that the Recorder Window UI seems to properly use the existing labels.
> Next steps that I'm expecting (maybe through further tasks/PRs):
> 
> * a quick UI review when the Recorder UX refactor is complete, to ensure everything is in place before documenting
> * a review of API docs for the public API items that were added through this PR

Think there is virtually no new public  API.
The manual might have to include a brief explanation of the  recorder integration.

PR #320, opened 2020-05-01
ABC-47 Recording alembic of objects that get destroyed at runtime stops recording

clusty:
The fix  comes from an user actually: https://github.com/Unity-Technologies/AlembicForUnity/pull/100/files

When new objects are discovered, there will be an ABC parent parent transform that is null.  
The fix is to climb up the hierarchy and reparent Abc nodes to match the GameObject hierarchy for orphan nodes.


com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## [2.2.0-exp.3] - 2021-01-22
 ### Changes
 - Minimum Unity version is 2019.4
+- Fixed a bug that caused the Alembic exporter to fail if GameObjects were being deleted during the recording session.
```
sebastienduverne, line 5: Description ok but the bullet should be in "Fixed" section.

sebastienduverne (commented):
Just a minor changelog structure issue to fix (see comment)

PR #326, opened 2020-05-11
Fixed OnValidate to clamp the minmaxslider properly when moving right cursor out of bounds in importer inspector

treque:
fix slider ftv 625


clusty:
Please change the base from master to dev ( edit the PR )

treque:
1- Should be OK now
2- Cannot reproduce on 2019.3.13f1
3- Christophe is creating another issue for this.

clusty:
Please add an entry in the changelog.

cguer:
Already spoke with Trang on Slack but adding notes here for tracking:
- Fix for #1 needs a bit more work (Trang is already on it)
- #2 in my previous comments can be ignored. Non-fixable after talking to Vlad.
- #3 in my previous comments can be ignored. It's out-of-scope for this PR, I'll open a separate JIRA ticket for i.

PR #327, opened 2020-05-12
Dev crash ftv 652

treque:
Added a null check so it stops crashing when entering play mode with a null meshfilter in mesh renderer


clusty:
Please add an entry in the changelog.

clusty:
Please give your PR a more human readable text :)

PR #328, opened 2020-05-14
FTV-597: Exporter SkinnedMesh scale bug

clusty:
When baking a skinned mesh, the world scale is ignored for <2020.2
In this case I add the scale myself. for 2020.2 there is a new flavour of the functions, where you can include the scale in the world transform.


com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
```diff
@@ -275,10 +273,42 @@ public void SetupSubmeshes(aeObject abc, Mesh mesh)
...
+                {
+                    var verts = new List<Vector3>();
+                    mesh.GetVertices(verts);
+                    for (var i = 0; i < verts.Count; ++i)
+                    {
+                        var v = verts[i];
+                        verts[i] = world2local.MultiplyPoint(v);
+                    }
+
+                    points.Assign(verts);
```
judubu, line 286: Can't you call LockList instead of Assign. It seems Assign is making a memcpy whereas LockList is avoiding the extra copy

PR #347, opened 2020-09-15
Convert the Burst dependency to a soft one.

clusty:
Burst created some issues for people trying to build with Mono x86_64 with a valid c++ compiler in their environment.
Switched to a conditional dependency.
This could impact slightly performance of the motion vector update.


clusty:
Will keep the hard dip, just update the version.

PR #348, opened 2020-09-15
Updated to Burst dependency 1.3.6

clusty:
Update to Burst 1.3.6 because 1.2 was causing issues with Mono builds when no C++ compiler was present on the client computer.


judubu (approved):
👍 +1:

PR #350, opened 2020-09-17
FTV-1122: Warn user of incompatible build target.

clusty:
Emit a warning when users download the Alembic package and build on unsupported platforms: Before they would get a runtime error if they actually used alembic.
Now we warn if there are alembics in the built scenes but allow the build to proceed.


com.unity.formats.alembic/Editor/Importer/AlembicBuildPostProcess.cs
```diff
@@ -21,6 +22,12 @@ public static void OnPostProcessBuild(BuildTarget target, string pathToBuiltProj
         {
             if (!TargetIsSupported(target))
             {
+                if (FilesToCopy.Count > 0)
+                {
+                    Debug.LogError(
+                        "Alembic does not support your current standalone build target. The currently supported targets are: Win64, OSX and Linux 64. ");
```
sebastienduverne, line 28: Suggestion:
"Alembic does not support your current standalone build target. You can currently choose only Windows 64-bit, MacOS X, or Linux 64-bit."
(+ remove extra space at the end of the string)

clusty: how does this sound:
**Alembic only supports the following standalone build targets: Windows 64-bit, MacOS X, or Linux 64-bit.**

Shorter, and does not give hope that new targets will be added :)

sebastienduverne: Right, in fact this was almost what I wrote for the doc :)
However, I checked in Unity's manual and it seems that `Standalone refers to builds for Mac, Windows and Linux.`
So I was wondering if we should remove "standalone" from here, because it could implicitly suggest that non-standalone targets are supported. Your opinion?

clusty:
We can.

judubu (approved):
👍

sebastienduverne (approved):
OK!

cguer (changes requested):
At what point of the build process should I see this warning? I tried a Windows (x86) and OpenGL builds but didn't get any warnings with either.

I tried:
- With just the Alembic package installed
- With an Alembic asset imported in the project
- With an Alembic GameObject instantiated in the scene

I also tried adding some random Logging statements around the code that was changed and nothing appears in the console 🤔

cguer (approved):
QA approved 👍!

Tested on:
- 2019.4.9f1 (Windows only)

PR #362, opened 2020-09-30
JIRA: ABC-164 Doc build warning

clusty:


com.unity.formats.alembic/Runtime/Scripts/Timeline/AlembicShotAsset.cs
```diff
@@ -37,7 +42,9 @@ public override Playable CreatePlayable(PlayableGraph graph, GameObject owner)
             return playable;
         }
 
-        /// <inheritdoc/>
+        /// <summary>
+        /// Return the duration in seconds of the Alembic asset.
```
sebastienduverne, line 46: Returns the duration of the Alembic asset, in seconds.
_**Note:** I'm not so familiar with the exact terminology to use to describe the involved thing, but I find "asset" a bit generic to associate a duration with it. Wouldn't we better use something like "Alembic playable asset"? (or whatever is properly describing the thing for which it makes sense to have a duration)._

sebastienduverne (commented):
See my inline comments.

sebastienduverne (approved):
LGTM

PR #368, opened 2020-11-05
Update lib alembic to 1.7.16

clusty:
The previous version of alembic 1.7.9 was 2 years old.
The main problem was that building it in debug failed on OS-X due to some c++17 incompatibilities.
Added build scripts for debug (alembic and 3rd party).


judubu (approved):
Good job. 
Note that you can pass the compiler flags directly to cmake. See my comment.

cguer (approved):
QA approved 👍!

Tested on:
- 2020.1.13f1 (Windows only)

PR #369, opened 2020-11-06
JIRA: ABC-182 Stadia Support

wackoisgod:
Adding support for Stadia (Stadia is support in 2019.3+ so I don't think we need to add any guards since that is the min version of the package.)


clusty (approved):
Thanks a lot for the PR.

judubu (approved):
👍

PR #372, opened 2020-11-20
Run unit tests on Ubuntu since it does not need GPU

clusty:
Run tests on ubuntu until the centOS image gets fixed to accept headless bokken VMs.
The GPU machine cause from time to time insane yamato wait times.

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -13,10 +13,15 @@ platforms:
     type: Unity::VM::osx
     image: package-ci/mac:latest
     flavor: m1.mac
-  - name: centOS
-    type: Unity::VM::GPU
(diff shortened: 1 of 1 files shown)


judubu (approved):
Approved but based on today's sync you'll need to rollback this one right?

clusty:
> Approved but based on today's sync you'll need to rollback this one right? Maybe :)

These tests can run without GPU. 
I added a new suite that can't, so if I run them all the time, the yamato pipeline is still stalled when GPU machines are scarce. Need to see if it's worth having a jungle of config options.

cguer (approved):
Nothing to QA but change is good 👍

clusty:
Out of date. Already merged in a different form.

PR #376, opened 2020-12-01
JIRA: ABC-175 Alembic binary libraries are copied into unsupported device builds.

clusty:
The 3 binary libs were copied into unsupported device builds: eg: iOS.

@cguer Pls run a standalone build on:
* OS-X confirm the Abc file plays fine.
* iOS there is no *abci* binary in the build


cguer (approved):
QA approved 👍!

Tested on:
- 2020.1.16f1 (MacOS only, build for iOS/MacOS)

PR #386, opened 2020-12-21
ABC-184 Nan Normals

clusty:
Bug was caused by bad triangles manufacturing 0 length normals, which when normalized create NaNs.

Fix was to check if norm is 0, return the original vector (crappy non-NaN normals).

@judubu All those changes are a cleanup of unused ISPC functions (since I can't debug, I had to blind fix the code.)
@cguer Please make me a lighter test asset to commit: I think an ill formed triangle can repro the bug.


cguer (approved):
QA approved 👍!

Tested on:
- 2020.2.0f1 (Windows only, HDRP only)

----

I wasn't able to reproduce the issue exactly as described in the [Fogbugz](https://fogbugz.unity3d.com/f/cases/1296269/) with a single degenerate triangle (ex: `[[0,0,0],[0,1,0,],[0,0,0]]`). It does indeed generate NaN normals but does not exhibit the "Bloom flickering issue". It appears we need at least one visible/rendered triangle for this.

Here is the test asset I ended up using. It has one degenerate triangle and one fully visible triangle (ex: `[[0,0,0],[0,1,0,],[0,0,0]]`, `[[0,0,0],[0,1,0,],[0,1,0]]`)
[trianglesNoWidth_noNormals_1f_24fps.zip](https://github.com/Unity-Technologies/AlembicForUnity/files/5787664/trianglesNoWidth_noNormals_1f_24fps.zip)

judubu (approved):

PR #389, opened 2021-01-04
ABC-173 Remove HDF5 support

clusty:
This PR drops the support for HDF5 (a very old obsolete alembic format).
This is needed because HDF5 comes as binary only distribution and blocks the port to Arm64 (Apple Silicon)


cguer (commented):
The change works as expected but I'm wondering if we could give our users a more explicit error if they try to import an HDF5. this is our current error:
Screenshot 2021-01-08 125357

----

I noticed that Blender uses the alembic file header to determine if it's an HDF5. I wonder if this is a reliable method we could also use at least to indicate our users the reason why Unity wasn't able to open their file?

https://github.com/blender/blender/blob/594f47ecd2d5367ca936cf6fc6ec8168c2b360d0/source/blender/io/alembic/intern/abc_reader_archive.cc#L47

cguer (approved):
QA approved 👍!

Tested on:
- 2020.2.0f1 (Windows only)

judubu (approved):
Yay!

redbosse:
Hello, does Unity Alembic support the Alembic format from Blender to export strands for hair

redbosse:
Hello, does Unity Alembic support the Alembic format from Blender to export strands for hair

PR #390, opened 2021-01-05
JIRA: ABC-180 Declare a dependency on cloth

clusty:
Starting end of February packages will be tested in true isolation: The generated test project will stop including any extra previously enabled packages.
Alembic had an undeclared dependency on the unity cloth module through the alembic exporter.

This Pr add this dependency explicitly.


cguer (approved):
QA approved 👍!

Tested on:
- 2020.2.0f1 (Windows only)

----

For general information: it is not possible to remove a module from the Package Manager Window when it is a dependency of another package. The following dialog appears:

Screenshot 2021-01-08 124846

It is still possible to remove a module by manually modifying the `manifest.json` but the restriction above gives us an extra safety net for most cases

clusty:
Of you remove from the manifest, it’s not really removed cause it would
break the dependencies.

judubu (approved):
Approved

PR #394, opened 2021-01-13
Repo readme

clusty:
First iteration on repo documentation.


clusty (commented):
I'm liking it :)

sebastienduverne:
I've put some more glue between the three .md files available in the repo's root to make the whole a bit more uniform, visible and part of the same user experience. However, I suppose we could have all on the same page. Opinions?

clusty:
> I've put some more glue between the three .md files available in the repo's root to make the whole a bit more uniform, visible and part of the same user experience. However, I suppose we could have all on the same page. Opinions?

I am a bit more reluctant to do that cause it might mess with the repo legal structure.

mfe (approved):
LGTM !
Feels as good as a spring cleaning ! :D

cguer (approved):
QA approved 👍!

PR #397, opened 2021-01-15
JIRA: ABC-100 Crash with file with empty mesh Samples

clusty:
This file is strange:
It contains empty meshes and Subdivision surfaces.

### DIFF
--- a/Source/abci/Importer/aiPolyMesh.cpp
+++ b/Source/abci/Importer/aiPolyMesh.cpp
@@ -338,12 +338,17 @@ void aiPolyMesh::updateSummary()
     // points
     {
         auto prop = m_schema.getPositionsProperty();
-        if (prop.valid() && prop.getNumSamples() > 0)
+        if (prop.valid() && prop.getNumSamples() > 0 )
         {
-            summary.has_points = true;
-            summary.constant_points = prop.isConstant();
-            if (!summary.constant_points)
-                m_constant = false;
+            Alembic::Util::Dimensions dim;
+            prop.getDimensions(dim);
+            if (dim.numPoints() > 0)
+            {
+                summary.has_points = true;
+                summary.constant_points = prop.isConstant();
+                if (!summary.constant_points)
+                    m_constant = false;
+            }
         }
     }
 
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Added support for arm64 macOS.
 - The package depends on the Cloth Unity Module.
 - Fixed a bug, where degenerate triangles would create NaN normals.
-- Fixed a crash caused by empty geometry samples.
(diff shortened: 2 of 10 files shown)


Source/abci/Importer/aiPolyMesh.cpp
```diff
@@ -338,12 +338,17 @@ void aiPolyMesh::updateSummary()
...
-        if (prop.valid() && prop.getNumSamples() > 0)
+        if (prop.valid() && prop.getNumSamples() > 0 )
         {
-            summary.has_points = true;
-            summary.constant_points = prop.isConstant();
-            if (!summary.constant_points)
-                m_constant = false;
+            Alembic::Util::Dimensions dim;
+            prop.getDimensions(dim);
+            if (dim.numPoints() > 0)
```
judubu, line 345: Shouldn't you do the same dimension check for every attributes?

com.unity.formats.alembic/Runtime/Scripts/Misc/Utils.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                array.Dispose();
+            }
+        }
+
+        public static NativeArray<T> ResizeIfNeeded<T>(this ref NativeArray<T> array, int newLength, Allocator a = Allocator.Persistent) where T : struct
+        {
+            if (array.Length != newLength)
+            {
+                array.DisposeIfPossible();
```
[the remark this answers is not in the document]
clusty: unfortunately this is an NativeArray and not a NativeList. Arrays have a fixed size at allocation.
This is merely a refactor to reuse the same "resizing" code instead of copy pasting it.

At some point I need to steal the NativeList code from unity.collections (dots). I cannot use it as is because they are not verified yet.

Source/abci/Importer/aiPolyMesh.cpp
clusty: good point. thought about it and then slipped my mind :)
Will do.

judubu: 👍

com.unity.formats.alembic/Runtime/Scripts/Misc/Utils.cs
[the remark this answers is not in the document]
judubu: Tanks for the explanation.

judubu (approved):
Ship it

cguer (approved):
QA approved 👍!

Tested on:
- 2020.2.0f1 (Windows only)

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Added support for arm64 macOS.
 - The package depends on the Cloth Unity Module.
 - Fixed a bug, where degenerate triangles would create NaN normals.
+- Fixed a crash caused by empty geometry samples.
```
sebastienduverne, line 10: Suggestion:
`Fixed the importer to prevent the Editor from crashing when importing meshes with empty geometry samples.`

PR #404, opened 2021-01-28
Third party notices fixes

clusty:
Remove HDF5 Boost from the thirdpartyNotices.
MurmurHash became public space according to the latest update to the Alembic lib:
```
Remove Murmur3 from the LICENSE.txt since it was moved into the public domain
a while ago, and is reflected as such in the code. (Issue 292)
```


mfe (approved):
LGTM ! Merci !

PR #405, opened 2021-02-01
JIRA: ABC-38 HDRP Alembic shader aren't use by default on import

clusty:
Alembic Mesh Default material support:
* legacy pipeline -> Alembic provided material with Motion Vectors
* HDRP -> Default Material (Lit) with custom velocity option
* URP -> Default ,material
This PR registers a custom dependency on ** GraphicsSettings.currentRenderPipeline** and retrievers an import if needed.

### DIFF
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Fixed
 - Fixed a bug that caused the Alembic exporter to fail if GameObjects were being deleted during the recording session.
 - Fixed a memory leak that occurred when using varying topology meshes.
-
+- Added a custom dependency on the "CurrentRenderPipeline" setting and assigning the correct SRP by default.
 ## [2.2.0-exp.2] - 2021-01-21
 ### Added
 - Added support for Stadia standalone builds.
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -5,7 +5,8 @@
 ### Fixed
 - Fixed a bug that caused the Alembic exporter to fail if GameObjects were being deleted during the recording session.
 - Fixed a memory leak that occurred when using varying topology meshes.
-- Added a custom dependency on the "CurrentRenderPipeline" setting and assigning the correct SRP by default.
+- Added a custom dependency on the "CurrentRenderPipeline" setting and assigning the correct SRP material by default.
+
 ## [2.2.0-exp.2] - 2021-01-21
 ### Added
 - Added support for Stadia standalone builds.
--- a/com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
+++ b/com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
@@ -250,10 +250,21 @@ namespace UnityEditor.Formats.Alembic.Importer
                 {
                     if (m_defaultMaterial == null)
                     {
-                        m_defaultMaterial = GetMaterial("Standard.shader");
+                        if (GraphicsSettings.currentRenderPipeline == null)
+                        {
+                            m_defaultMaterial = GetMaterial("Standard.shader");
+#if HDRP_PRESENT
+
+#endif
+                        }
+                        else
+                        {
+                            m_defaultMaterial = Instantiate(GraphicsSettings.currentRenderPipeline.defaultMaterial);
+                        }
+
+                        Add("Default Material", m_defaultMaterial);
                         m_defaultMaterial.hideFlags = HideFlags.NotEditable;
                         m_defaultMaterial.name = "Default Material";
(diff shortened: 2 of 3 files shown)


com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
```diff
@@ -203,6 +211,27 @@ public override void OnImportAsset(AssetImportContext ctx)
             firstImport = false;
         }
 
+        [InitializeOnLoadMethod]
+        static void InitializeEditorCallback()
+        {
+            EditorApplication.update += DirtyCustomDependencies; //
```
bastienunity, line 217: You should call DirtyCustomDependencies once here so it has a chance to setup the dependency value before the first import of the project assets starts.
This would prevent the double import of the alembic assets when opening a project after updating Unity or with the Library folder removed.

com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
```diff
@@ -203,6 +211,27 @@ public override void OnImportAsset(AssetImportContext ctx)
...
+        }
+
+        static string pipelineName = string.Empty;
+        static void DirtyCustomDependencies()
+        {
+            var newPipelineName = GraphicsSettings.currentRenderPipeline == null
+                ? ""
+                : GraphicsSettings.currentRenderPipeline.ToString();
```
bastienunity, line 225: Use AssetDatabase.TryGetGUIDAndLocalFileIdentifier instead of the name and compute the hash from the guid and localId.
This will handle multiple pipelines with the same names and multiple pipelines declared in one configuration asset (yes, we don't do it usually, but that's a doable thing that needs to be taken into account here for determinism)

com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
```diff
@@ -138,13 +141,17 @@ void OnValidate()
...
+        internal const string renderPipepineDependency = "AlembicRenderPipelineDependency";
+
         public override void OnImportAsset(AssetImportContext ctx)
         {
             if (ctx == null)
             {
                 return;
             }
 
+            ctx.DependsOnCustomDependency(renderPipepineDependency);
```
bastienunity, line 153: Do you want to always depend on the custom render pipeline?
Is it possible a model doesn't have any material created from the default material and thus doesn't have to re-import on pipeline change?

bastienunity (changes requested):
One general question and two changes requested.
This is a good step to determinism for this importer though :)

bastienunity (approved):
Looks good to me now!

judubu (approved):
:+1: 
I'll keep that in mind for a future USD work

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## [2.2.0-exp.3] - 2021-01-22
 ### Changes
 - Minimum Unity version is 2019.4
+- Added a custom dependency on the "CurrentRenderPipeline" setting and assigning the correct SRP by default.
```
sebastienduverne, line 5: Suggestion:
`- Added a custom dependency on the "CurrentRenderPipeline" setting and ensured to assign the correct SRP by default.`

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## [2.2.0-exp.3] - 2021-01-22
 ### Changes
 - Minimum Unity version is 2019.4
```
sebastienduverne, line 4: Since we're here:
`- Changed the Unity Editor minimum compatible version to 2019.4.`

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## [2.2.0-exp.3] - 2021-01-22
 ### Changes
```
sebastienduverne, line 3: `Changes` => `Changed`
I know we didn't officially state on the changelog section names, but we now tend to uniformize with these (in this order):
- Added
- Changed
- Fixed

sebastienduverne (commented):
Changelog overall review comments

clusty:
@cguer Please test these cases at least:
Legacy Pipeline: Same thing as now
HDRP: grey lit shader with motion vectors
URP: default shader (never tried it myself :) )

cguer (approved):
QA approved 👍!

Tested on:
- 2020.2.4f1 (Windows only)

Validated:
- Clean import in new project (HDRP, URP, Legacy)
- Changing SRP while project is open. Objects with default materials are correctly updated (and objects with explicit material assignments are untouched)
- Asset import previews now display correct materials with all SRPs
- Upgrading from a previous version of the Alembic packages properly updates materials

----

I almost cried tears of joy reviewing this, it truly is a beautiful day
giphy

PR #406, opened 2021-02-02
JIRA: ABC-105 Exported normals can't be read outside of Unity

clusty:
The bug was caused by the fact that the normals did not have the correct scope. All vertex attributes suffer from this issue:
UV1, UV2, Colours, Normals.

If number of Indices is the same as the number of faces * 3 (every point is used in only 1 face ), attributes are FaceVarying, else they are Vertex scope

### DIFF
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 - Fixed a bug that caused the Alembic exporter to fail if GameObjects were being deleted during the recording session.
 - Fixed a memory leak that occurred when using varying topology meshes.
+- Fixed a bug that caused the exported mesh normals to be unreadable in DCCs.
 
 ## [2.2.0-exp.2] - 2021-01-21
 ### Added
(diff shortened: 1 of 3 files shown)


com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## [2.2.0-exp.3] - 2021-01-22
 ### Changes
 - Minimum Unity version is 2019.4
+- Fixed a bug that caused the exported mesh normals to be unreadable in DCCs
```
sebastienduverne, line 5: Just move to "Fixed" section + add a period.

sebastienduverne (commented):
Minor changelog structure issue.

clusty:
@cguer Please test that Maya is not complaining about exported mesh attributes: bonus points if whacky unity normals are preserved (could manufacture some normals by code if need be...).

cguer (approved):
QA approved 👍!

Tested on:
- 2020.2.3f1 (Windows only)

Validated:
- No attribute warning when importing alembic file into Maya
- "Messed up" normals in Unity are properly exported
- Works for tri and quad meshes, as well as sub-meshes with no connected vertices

(unrelated to this PR but also confirmed that quad polygons are preserved on export)

judubu (approved):
👍

PR #407, opened 2021-02-03
JIRA: ABC-191 Add import option to add the AlembicCurveRenderer automatically

clusty:
Screen Shot 2021-02-03 at 9 00 51 AM

This PR adds a new option called "**AddCurveRenderers**" which adds automatically the AlembicCurveRenderer components on every AlembicCurve.

This is useful when one desires just a visualization of the curves (or they don't have access to fancier curve rendering)

This PR affects the package user guide => to be managed through [ABC-189](https://jira.unity3d.com/browse/ABC-189).


com.unity.formats.alembic/Editor/Importer/AlembicImporterEditor.cs
```diff
@@ -48,8 +48,13 @@ public override void OnInspectorGUI()
...
                 EditorGUILayout.PropertyField(serializedObject.FindProperty(pathSettings + "importPoints"),
                     new GUIContent("Import Points", ""));
-                EditorGUILayout.PropertyField(serializedObject.FindProperty(pathSettings + "importCurves"),
-                    new GUIContent("Import Curves", ""));
+                var importCurvesProp = serializedObject.FindProperty(pathSettings + "importCurves");
+                EditorGUILayout.PropertyField(importCurvesProp, new GUIContent("Import Curves", ""));
+                if (importCurvesProp.boolValue == true)
+                {
+                    EditorGUILayout.PropertyField(serializedObject.FindProperty(pathSettings + "createCurveRenderers"),
+                        new GUIContent("Add Curve Renderers", ""));
```
cguer, line 56: Suggestion for tooltip: `Automatically add "Alembic Curves Renderer" components on curve objects. This component enables basic rendering of curves for preview purposes inside the Game and Scene views.`

I think we need to mention something like "preview" or "visualize" to make sure that users don't assume a full rendering solution (which I think some will).

cguer (commented):
The feature is QA approved but I think a tooltip would be needed for the option. I added a suggestion in-line, let me know what you think.

mfe (approved):
LGTM, but second Christophe on the tooltip.

com.unity.formats.alembic/Editor/Importer/AlembicImporterEditor.cs
clusty: @sebastienduverne how does this sound?
*Automatically add "AlembicCurvesRenderer" components on curve objects. This component enables basic preview of curves.*

judubu (approved):
👍

com.unity.formats.alembic/Editor/Importer/AlembicImporterEditor.cs
sebastienduverne: I would just change the 2nd sentence:
`This allows you to preview curves in low resolution.`

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicSettings.cs
```diff
@@ -174,5 +174,15 @@ public bool ImportCurves
             get { return importCurves; }
             set { importCurves = value; }
         }
+
+        [SerializeField] bool createCurveRenderers;
+        /// <summary>
+        /// If importing of Alembic curves is enabled, also create automatically the AlembicCurveRendering component.
```
sebastienduverne, line 180: Suggestion:
`If you enabled the importing of Alembic curves, this method automatically creates the AlembicCurveRendering component.`

sebastienduverne (commented):
See my in-line comments for changelog, tooltip and API doc.

com.unity.formats.alembic/Editor/Importer/AlembicImporterEditor.cs
cguer: The preview isn't actually lower resolution. The curves will get rendered at the resolution of the scene/game view like all other objects in the scene.

Without this component though, the curves are never visible in the scene/game view. The main purpose of the component is to make the curves visible in the scene/game view but it shouldn't be used for "final quality render".

What makes this a "preview" is the limited quality/options for rendering. For example, a key feature of curves is the ability to modulate its width ([example](https://docs.arnoldrenderer.com/display/A5AFMUG/Curves#Curves-CurveWidth#Curves-CurveWidth)) but this is not possible with `AlembicCurvesRenderer` alone.

If a user wanted to control the width, they would have to code their own rendering solution (converting the curve data into visual information on screen). The demo team is also currently working on a bigger workflow (which includes Alembic as an importer) with more advanced rendering features. This is likely what users will be using in the future to get "good" curve renders inside Unity.

sebastienduverne: Thanks for the clarification. Take 2 (2nd sentence only):
`This allows you to get a basic preview of the Alembic curves in the Scene.`

As for the curve features not available, this is typically nice details to have in documentation, I'll keep it in mind.

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changes in Alembic for Unity
 ## [2.2.0-exp.3] - 2021-01-22
+### Added
+- Added an option to automatically add the AlembicCurveRendering components.
```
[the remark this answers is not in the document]
sebastienduverne: New suggestion here too:
`New option to automatically add the AlembicCurveRendering components for basic preview of the curves in the Scene.`

sebastienduverne (approved):
LGTM

cguer (approved):
QA approved 👍!

PR #410, opened 2021-03-10
JIRA: ABC-201 Memory leak in the Chase Cinematic template

clusty:
Bug was caused by having multiple timeline tracks that enable/disable/update the same AlembicStreamPlayer on the same frame. Because of that the same stream would be reopened without destroying the old stream.

@cguer My repro steps:
Play the chase demo once from the beginning to the end. 
Exit playmode
Replay the chase and notice the console errors
Please also check https://jira.unity3d.com/browse/ABC-90 : I unable to repro with my fix and cause seems to be related.

### DIFF
--- a/com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStreamPlayer.cs
+++ b/com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStreamPlayer.cs
@@ -182,6 +182,11 @@ namespace UnityEngine.Formats.Alembic.Importer
         {
             if (StreamDescriptor == null)
                 return false;
+            if (abcStream != null)
+            {
+                abcStream.Dispose();
+            }
+
             abcStream = new AlembicStream(gameObject, StreamDescriptor);
             var ret = abcStream.AbcLoad(createMissingNodes, false);
             forceUpdate = true;
@@ -257,7 +262,11 @@ namespace UnityEngine.Formats.Alembic.Importer
         void OnDisable()
         {
             if (abcStream != null)
+            {
                 abcStream.Dispose();
+            }
+
+            abcStream = null;
         }
 
         void OnApplicationQuit()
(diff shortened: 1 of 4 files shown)


cguer (approved):
QA approved 👍!

Tested on:
- 2020.3.0f1 (Windows only)

----

Also confirmed that this PR **fixes** the activation bug for https://jira.unity3d.com/browse/ABC-90

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStreamPlayer.cs
```diff
@@ -257,7 +262,11 @@ void OnEnable()
         void OnDisable()
         {
             if (abcStream != null)
+            {
                 abcStream.Dispose();
+            }
+
+            abcStream = null;
```
judubu, line 269: It would make more sense to move it inside the brackets and set it only when its not already null.

com.unity.formats.alembic/Tests/Editor/EditorTests.cs
```diff
@@ -54,8 +58,43 @@ public IEnumerator AddCurveRenderingSettingIsObeyed([Values] bool addCurve)
...
+            c2.duration = 1;
+
+            director.SetGenericBinding(a1, player.gameObject);
+            director.SetGenericBinding(a2, player.gameObject);
+            director.RebuildGraph();
+            director.time = 0;
+            yield return null;
+            director.time = 2.5;
+            director.Evaluate();
+            yield return null;
```
judubu, line 97: I don't see any assertion. Are you just expecting the thing not to produce errors?

com.unity.formats.alembic/Tests/Editor/Unity.Formats.Alembic.Editor.Tests.asmdef
```diff
@@ -1,14 +1,25 @@
 {
```
judubu, line 1: Are those all auto generated by Unity?

judubu (approved):
Couple of minor comments but it looks good to me.

PR #415, opened 2021-03-30
JIRA: ABC-206 "Add Curve Renderers" does not automatically add component for import-less Alembic

clusty:
It was an UI issue that the settings were not flushed before the update was called.
@cguer : This affects the import less stream settings and undo.


cguer (approved):
QA approved 👍!

Tested on:
- 2020.3.0f1 (Windows only)

judubu (approved):
👍

PR #422, opened 2021-04-19
Added support for not reloading materials.

dgsantana:
This would help when loading abc from a network location at runtime.

### DIFF
--- a/com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStreamPlayer.cs
+++ b/com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStreamPlayer.cs
@@ -171,7 +171,7 @@ namespace UnityEngine.Formats.Alembic.Importer
         /// </summary>
         /// <param name="newPath">Path to the new file.</param>
         /// <returns>True if the load succeeded, false otherwise.</returns>
-        public bool LoadFromFile(string newPath)
+        public bool LoadFromFile(string newPath, bool reloadMaterials = true)
         {
             if (StreamDescriptor == null)
             {
@@ -179,7 +179,7 @@ namespace UnityEngine.Formats.Alembic.Importer
             }
 
             StreamDescriptor.PathToAbc = newPath;
-            return InitializeAfterLoad();
+            return InitializeAfterLoad(reloadMaterials);
         }
 
         /// <summary>
@@ -230,7 +230,7 @@ namespace UnityEngine.Formats.Alembic.Importer
             }
         }
 
-        bool InitializeAfterLoad()
+        bool InitializeAfterLoad(bool reloadMaterials = true)
         {
             var ret = LoadStream(true, true);
             if (!ret)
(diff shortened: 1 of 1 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/AlembicForUnity?pullRequest=422) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/AlembicForUnity?pullRequest=422) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/AlembicForUnity?pullRequest=422) it.</sub>

clusty:
Hey,

Thanks for your PR.
Mind giving some info on what problem are you trying to solve ?

dgsantana:
Sure. We have a case where we need to keep the abc files outside the project and also outside the builds (not in the streaming assets). For keeping the abc files outside the project we use addressables, and dynamically load the addressable either at play mode time or in a build. We need to reload the abc file with the correct path, but keeping any material changes we make. With that small change, it's possible to do that, by just passing false in the reloadMaterials optional parameter.
I think this would help other that want to keep the abc files outside the project using addressables or other system.

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStreamPlayer.cs
```diff
@@ -243,20 +243,22 @@ bool InitializeAfterLoad()
...
-
-            foreach (var meshRenderer in gameObject.GetComponentsInChildren<MeshRenderer>(true))
+            if (reloadMaterials)
             {
-                var mats = meshRenderer.sharedMaterials;
-                for (var i = 0; i < mats.Length; ++i)
+                var defaultMat = AlembicMesh.GetDefaultMaterial();
+                foreach (var meshRenderer in gameObject.GetComponentsInChildren<MeshRenderer>(true))
                 {
-                    if (mats[i] == null) // Add the default material if no material present
```
clusty, line 253: This code changed in the last 2-3 weeks.
Are you sure you still need your changes ?
Basically this null check here assures that only null material references are replaced with default materials (this is to prevent default magenta materials).

dgsantana: Ok, indeed it should also solve the issue. The latest official package didn't had that so I didn't saw that when I was making the changes, but seems a much more simple and effective way. I will test the dev branch tomorrow on our codebase and see if it works. Thanks @clusty.

dgsantana: Did a quick test, and it's working. When do you expect that this will be available on the package manager as either preview or release?

This PR can be close, no need for any change.

clusty:
I can’t talk specific dates, but expect it soon.
Please reach back here or on the forum if the issue is still blocking.

PR #423, opened 2021-04-19
JIRA: ABC-214 Entering Playmode with Alembic inspector in focus throws error with SRP

clusty:
Entering playmode, would trigger a reimport since dependencies change.


judubu (approved):

cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.0a12 (Windows only)

PR #427, opened 2021-04-27
JIRA: ABC-217 Import-less doesn't create enough material slots for all submeshes

clusty:
Easy peasy: create enough materials slots for every sub mesh.


com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStreamPlayer.cs
```diff
@@ -245,19 +245,30 @@ bool InitializeAfterLoad()
 
             var defaultMat = AlembicMesh.GetDefaultMaterial();
 
-            foreach (var meshRenderer in gameObject.GetComponentsInChildren<MeshRenderer>(true))
+            gameObject.DepthFirstVisitor(go =>
```
judubu, line 248: Neat!

judubu (approved):
👍

PR #428, opened 2021-04-30
JIRA: ABC-56 Crash with variable topology file

clusty:
The bug appears because the file contains samples with no points in the middle of the file.
This is a punctual fix for this problem, but all other attributes could have this issue (or mismatched array sizes).
In case this becomes a more widespread problem, a deeper validation needs to happen for the whole data.


com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 ### Changed
 ### Fixed
+- Fixed a bug that caused out of project stream files to have the wrong number of material slots.
```
clusty, line 11: @sebastienduverne this is previous merge's changelog entry.

sebastienduverne: Suggestion:
`- Fixed a bug that caused files streamed from outside the project to have the wrong number of material slots.`

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Fixed a bug that caused crashed when the file was of variable topology and there were empty vertex samples.
```
sebastienduverne, line 12: Suggestion:
`- Fixed a bug that was leading the Editor to crash when importing a file with variable topology and empty vertex samples.`

sebastienduverne (approved):
Changelog OK.

cguer (changes requested):
The Editor doesn't crash anymore but I get these errors when playing back (or scrubbing the `Time` property) with the problematic Alembic file:
Screenshot 2021-05-05 102141

```
Invalid worldAABB. Object is too large or too far away from the origin.
```
```
Mesh 'dyn: beamShape': abnormal mesh bounds - most likely it has some invalid vertices (+/-inifinity or NANs) due to errors exporting.
Mesh bounds min=(-432003559175647211775943811115515904.00, -1740458287276054289552242818565537792.00, -1714777904503980761279372230855229440.00), max=(424979863726513906851769749602304.00, 51924553148598561572056834171207680.00, 31168359133111590409300190062182400.00). Please make sure the mesh is exported without any errors.
UnityEngine.GUIUtility:ProcessEvent (int,intptr,bool&)
```
```
In order to call GetTransformInfoExpectUpToDate, RendererUpdateManager.UpdateAll must be called first.
UnityEngine.GUIUtility:ProcessEvent (int,intptr,bool&)
```
```
Assertion failed on expression: 'IsFinite(outDistanceForSort)'
UnityEngine.GUIUtility:ProcessEvent (int,intptr,bool&)
```

PR #432, opened 2021-05-05
JIRA: ABC-222 Alembic Reimports on on every SRP project start

clusty:
The bug was caused by the fact that when the editor restarts the pipeline dependency is not consistent with the current value.

### DIFF
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 ### Fixed
 - Fixed a bug that was converting previous Alembic instance in prefabs into out-of-project references.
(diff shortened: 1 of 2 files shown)


com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 ### Fixed
 - Fixed a bug that was converting previous Alembic instance in prefabs into out-of-project references.
+- Fixed a bug that was causing Alembic files to be re-imported on every SRP project start.
```
sebastienduverne, line 12: Suggestion:
`- Fixed a bug that was re-importing Alembic files on every SRP project start.`

sebastienduverne (approved):
Changelog OK!

cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.0a15 (Windows only, URP and HDRP)

judubu (approved):

PR #433, opened 2021-05-05
[ABC-223] Standalone Alembic Builds fail when building twice with read only assets 

clusty:
Overwrite file.copy arg does not work on readonly files on Windows.
Also added a small optimization to not copy the same file multiple times (multiple instances of the same file.)


judubu (approved):
👍

sebastienduverne (approved):
Changelog OK as is.

cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.0a15 (Windows only)

PR #443, opened 2021-06-01
Support for reading ISubD Schema meshes (ABC-159)

cpredmann:
This PR is to handle reading of objects with an ISubD schema

- Moves mesh handling logic into a templated class to be used with both ISubD and IPolyMesh
- Normals do not exist on ISubD, so always calculates normals
- Exported meshes are still written as IPolyMesh and no crease information is preserved


com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicSubD.cs
```diff
@@ -0,0 +1,22 @@
...
+        aiSubD m_abcSchema;
+
+        internal override aiSchema abcSchema
+        {
+            get { return m_abcSchema; }
+        }
+
+        internal override void AbcSetup(aiObject abcObj, aiSchema abcSchema)
+        {
+            base.AbcSetup(abcObj, abcSchema);
```
clusty, line 16: Do think this is needed ?

Larger picture: Is there any value in having a separate AlembicNode type in C# ? Could we have both polyMesh and subD c++ classes always create the same C# PolyMesh type ?
SubD is nothing more than a mesh today

cpredmann: That's a good point.  I thought about that, and didn't know if it was worth having something as a placeholder for future attribute possibilities (crease weights, etc.).  But, might just be confusing since it doesn't do anything more than a Mesh.

clusty: You might be right.
Let's leave it as is for now: Maybe the adaptive GPU Subdividing project will catch on and there'll be interest for creasing info.

clusty (approved):

cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.0b2 (Windows only)

judubu (approved):
GG. Ship it.

PR #444, opened 2021-06-03
JIRA: ABC-224 Alembic Motion vectors persist when the model is not being updated

clusty:
Before:
If a frame included a time change update the mesh including velocity
If a frame had no time change early out, causing motion vectors to persist

after:
If a frame included a time change update the mesh including velocity
If a frame had no time update, zero out the mesh velocity once.

@cguer 
Compare the velocities inside HDRP of a cube that moves and then stops in the middle of a screen with the same skinned cube recorded as alembic.

### DIFF
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 ### Changed
 ### Fixed
-- Fixed a bug that caused visual artefacts when pausing Alembic playback. 
+- Fixed a bug that was causing visual artefacts due to motion vector persistence when pausing Alembic playback.
 
 ## [2.2.0-pre.4] - 2021-04-27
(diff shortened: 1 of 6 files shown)


cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.0a17 (Windows only)

judubu (approved):
LGTM

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 ### Changed
 ### Fixed
+- Fixed a bug that caused visual artefacts when pausing Alembic playback.
```
sebastienduverne, line 11: Suggestion, where we could probably include a slice of the cause, which explicitly involves motion vectors (?):
`Fixed a bug that was causing visual artefacts due to motion vector persistence when pausing Alembic playback.`

sebastienduverne (approved):

PR #451, opened 2021-06-21
JIRA: ABC-171 Automatically resize Alembic clips to the animation length on creation

clusty:
When dragging an alembic stream player onto a timeline alembic track, the length of the clip should be the same as a the StreamPlayer's duration:

@cguer strange cases to test:
* Duplicate a clip should preserve source length.
* Mess with clip in Timeline asset mode 
* Drag a clip that has only 1 frame
* Test mechanism when inside subtimelines.
* Importerless StreamPlayer

### DIFF
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for importing the cage mesh of SubD Alembic nodes.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


cguer (changes requested):
Already discussed with Vlad in person but adding here for history:

Everything looks good to me except when the Alembic file only has 1 sample. It creates a tiny clip that is hard to manipulate and we can assume such "static" meshes would usually be used for longer.
To avoid this, we'll fallback to the default clip length (5s) if the Alembic file only has one frame/sample.

cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.0b2 (Windows only)

sebastienduverne (approved):
Changelog OK

judubu (approved):

PR #456, opened 2021-07-09
JIRA: ABC-154 Implement facesetNames and importer UI

clusty:
This PR implement material remapping for the alembic Importer.
There is a new UI tab called materials where all faceset names are shown, grouped by Mesh
Materials can be overridden by hand by populating the field, or can be auto-assigned (Chose search location and press auto-assign).
When auto-assigning, the first material with the same name as the faceset name is chosen.

@cguer Please test things around the following cases:
* Partial assignment of materials
* Undo/Redo's of assignments
* Upgrade a previous version of a file with/without default materials (prefab instances)

PR

The same faceset data is available in the **AlembicCustomData** componets on relevant gameObjects, for custom user assignment strategies.


judubu:
Starting with the UI:
* no indentation after the root is surprising, it makes me wonder why there's no material box on the right of the root. Maybe you can get rid of the root altogether given the name is already present in the importer settings name.
* it could be useful to have the full alembic path in the tooltip of each mesh, in the case you have duplicate names

judubu (approved):
Great work.

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [2.3.0-pre.1] - 2021-07-11
 ### Added
 - Added support for importing the cage mesh of SubD Alembic nodes.
+- Added support for reading faceset names and Importer material assignment UI tools.
```
sebastienduverne, line 10: I don't get "Importer material assignment UI tools".

cguer: I'm not sure the best way to phrase it but maybe the second part of the sentence could be on a different line. Something  like:
`- Added tool in the importer to automatically assign materials based on faceset names.`

cguer (approved):
QA approved 👍!

Tested on:
- 2021.1.15f1 (Windows only)

sebastienduverne (commented):
- Made a suggestion for Changelog.
- Question: the "correct" way to mention _Face Sets_ seems to be as I'm writing it here. Do you think we should/could update it in the UI? (I started a thread about this in our Alembic internal Slack channel).

com.unity.formats.alembic/CHANGELOG.md
clusty: Fixed.

clusty:
> * Made a suggestion for Changelog.
> * Question: the "correct" way to mention _Face Sets_ seems to be as I'm writing it here. Do you think we should/could update it in the UI? (I started a thread about this in our Alembic internal Slack channel).

Updated the UI. Here are the places I saw the UI showed Faceset
Screen Shot 2021-08-12 at 2 13 33 PM
Screen Shot 2021-08-12 at 9 19 47 AM

sebastienduverne (approved):
LGTM

com.unity.formats.alembic/Runtime/Scripts/Importer/Behaviours/AlembicCustomData.cs
```diff
@@ -0,0 +1,27 @@
+using System.Collections.Generic;
+
+namespace UnityEngine.Formats.Alembic.Importer
+{
+    /// <summary>
+    /// Class that stores ancillary data for the current Alembic Node.
```
sebastienduverne, line 6: I didn't even know the word "ancillary" existed (learning something every day 😸).
Although I think it's correct, I would probably use more generic terminology such as "additional", for example.

com.unity.formats.alembic/Runtime/Scripts/Importer/Behaviours/AlembicCustomData.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    public class AlembicCustomData : MonoBehaviour
+    {
+        [SerializeField]
+        List<string> faceSetNames;
+
+        /// <summary>
+        /// Retrieves the set of Face Set names stored in the current Alembic node.
```
sebastienduverne, line 14: I understand this is a set, but as we unfortunately have the same term in Face Set, it makes the sentence a bit unclear.
I would go direct with "Retrieves the Face Set names..."

sebastienduverne (approved):
Just two minor suggestions/corrections, but overall API docs LGTM.

com.unity.formats.alembic/Runtime/Scripts/Importer/Behaviours/AlembicCustomData.cs
```diff
@@ -3,15 +3,15 @@
...
+    /// Class that stores additional data for the current Alembic Node.
     /// </summary>
     public class AlembicCustomData : MonoBehaviour
     {
         [SerializeField]
         List<string> faceSetNames;
 
         /// <summary>
-        /// Retrieves the set of Face Set names stored in the current Alembic node.
+        /// Retrieves the set of Face Set names.
```
sebastienduverne, line 14: I was rather suggesting:
`Retrieves the Face Set names stored in the current Alembic node.`

PR #457, opened 2021-07-09
Replace raw vector

clusty:
This is a pre-requisite refactor for a crash / memory coruption bug ([related bug PR](https://github.com/Unity-Technologies/com.unity.formats.alembic/pull/461), [ABC-76](https://jira.unity3d.com/browse/ABC-76)).
The RawVector implementation is unsafe: in debug bounds are not checked leading to memory corruption.
Also empirical tests seem to make STL Vectors slightly faster.
RawVectors used a custom allocations that were 32bit aligned (for SIMD math). on 64bit Architectures all STL vectors are guaranteed to be 32 bit aligned as well.

@cguer Pls check no regressions appear: 
* Try using: mesh (with/without vertex colours), point and curve data.
* Please do your tests on Win/OSX


sebastienduverne (approved):

Source/abci/Foundation/Vector.h
```diff
@@ -0,0 +1,96 @@
+#pragma once
+#include<vector>
+
+template <class T>
+using Vector = std::vector<T >;
+
+template <class T>
+void CopyTo(const Vector<T> &src, T* dst)
+{
+    memcpy(dst, src.data(), sizeof(T) * src.size());
```
judubu, line 10: No need to guard against dst == nullptr?

Source/abci/Foundation/Vector.h
```diff
@@ -0,0 +1,96 @@
...
+}
+
+template <class T>
+void CopyTo(const Vector<T> &src, T* dst, size_t length, size_t offset = 0)
+{
+    memcpy(dst, src.data() + offset, sizeof(T) * length);
+}
+
+template<class T, class U>
+void Assign(Vector<T>& dst, const U& src, int point_count)
```
judubu, line 20: I'd rename point_count to just count or count in the whole file

judubu (commented):
Couple of minor comments in Vector.h

Source/abci/Foundation/Vector.h
clusty: Good point.

judubu (approved):

cguer (approved):
QA approved 👍! Clicked on all the things and couldn't find anything broken.

Tested on:
- 2021.1.15f1 (Windows, MacOS)

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.formats.alembic?pullRequest=457) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.formats.alembic?pullRequest=457) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.formats.alembic?pullRequest=457) it.</sub>

clusty:
> I just noticed this PR. I want to explain why I added RawVector. I don't intend to stop this change, but I just want you to know. there were two reasons:
> 
> 1. for faster resize(). std::vector::resize() initialize newly added elements. POD types like float3 are zero cleared. zero clear for huge vertex arrays is surprisingly slow. I realized the fact by profiling with VTune, and it was the big reason to add RawVector.
> 2. for 32 byte align. malloc()ed memory is guaranteed to 16 byte aligned , but not guaranteed to 32 byte aligned (on Windows).
>    aiSIMD.ispc uses AVX-512 if available and in that case, non 32 byte aligned memory causes a crash.
>    I think this is still the case, but this problem can be easily avoided by removing avx512skx-i32x16 target (CPUs with AVX-512 are quite rare) or adding a custom aligned allocator for std::vector.

Hey,

Unfortunately this PR is agonizing for a very long time so who knows when it will get merged....
I am encountering a a bunch of crashes with RawVectors because they write out of bounds and this makes it very hard to debug (the stl vector is checked for bounds I think when building in debug).

Empirical checks have shown that the StlVector is a tiny bit faster compared to RawVector (granted I checked only briefly on my OSX)

Thanks a lot for the heads up regarding alignment: I have a buggy aligned allocator already implemented but did not push it.
I thought I read that "int" is always 4 byte aligned on 64bit systems.

lucillecaillaud:
We have no intention to put more effort into this for now. This PR is linked in the related JIRA [ABC-76](https://jira.unity3d.com/browse/ABC-76) in the comment and can be re-opened when appropriate.

PR #458, opened 2021-07-09
JIRA: ABC-94 World scaling ignored on Unity cloth export

clusty:
The world scale was ignored by the cloth exporter (Similar to the previous skinned mesh issue.)

Deleted the unit test cause I could not find a way to make the Cloth simulation deterministic

### DIFF
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 ### Fixed
+- Fixed a bug that caused Unity Cloth alembic export to ignore the world scale.
 
 ## [2.2.0] - 2021-06-10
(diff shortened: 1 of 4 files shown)


judubu (approved):

cguer (approved):
QA approved 👍!

Tested on:
- 2021.1.15f1 (Windows only)

-----

Regarding the stability of cloth sim for unit tests, could you disable the gravity? That way, the mesh is still a "cloth" but should not move at all since there are no physics.
cloth_gravity

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 ### Fixed
+- Fixed a bug that caused Unity Cloth alembic export to ignore the world scale.
```
clusty, line 13: "A"lembic

PR #459, opened 2021-07-12
JIRA: ABC-137 Empty GameObjects ignored from export

clusty:
The transforms were exported only as a side effect of cameras / meshes, etc.
Now pure tranforms animations are exported as well.


judubu (approved):

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for importing the cage mesh of SubD Alembic nodes.
 
 ### Changed
+- Alembic now exported GameObjects now inclue also GameObjects that change only their transform.
```
clusty, line 12: Alembic now exports transform hierarchies when child GameObjects don't contain meshes.

cguer (commented):
Exporting a single gameobject with only a Transform component is generating two XForm nodes in the Alembic file. Is this the expected behavior?

Output from `abcinfo`:
```
D:\unity\qa\2021.1.15f1_3d\Assets\Scenes\Alembic Clip.abc, version: 10716 (2 timings)
   TimeSample[0]      1 samples:  t0=0 (frame 0.0 @ 24fps)
   TimeSample[1]      1 samples:  t0=0 (frame 0.0 @ 24fps)

Objects:
/emptyGameobject (FFFFC094) (Xform)
/emptyGameobject (FFFFC094)/emptyGameobject (FFFFC094) (Xform)
```

Also, **both** Xform nodes have the transforms of the original gameobject. Screenshot in Unity of the object getting exported and what importing the resulting alembic file looks like:
Screenshot 2021-07-15 131602

clusty:
@cguer I fixed the double transform issue.

Would you please check that exporting sub-hierarchies of meshes works as before ? (try with non identity parents on top of exported meshes)

cguer (approved):
QA approved 👍!

Tested on:
- 2021.1.15f1 (Windows only)

PR #461, opened 2021-07-14
JIRA: ABC-76 Empty GameObjects are created when applying changed properties of imported Alembic file with a missing Imported Object

clusty:
This bug was caused by importing an Alembic file that contained nodes with non-standard schemas. The error was appearing when these nodes contained children.

After the fix, unrecognized schema nodes create NOOP GameObjects and a warning is printed in the console.


com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
```diff
@@ -383,6 +382,10 @@ void ImportCallback(aiObject obj)
             else
             {
                 obj.SetEnabled(false);
+                if (ic.createMissingNodes)
+                {
+                    Debug.LogWarning($"{childName} has an unsupported schema.");
```
clusty, line 387: @sebastienduverne pls validate as well.

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for importing the cage mesh of SubD Alembic nodes.
 
 ### Changed
+- Unsupported schema nodes are imported as NOOP GameObjects.
```
clusty, line 12: Alembic importer now imports unsupported schema nodes as empty GameObjects.

judubu: no-op maybe? NOOP reads like a type of game objects

judubu (approved):
👍

cguer (changes requested):
The issue is fixed for the problematic file (`sfm.abc`) but now facesets are also recognized as unsupported and log warnings.

Example when importing this file [michl_simpleMaterials_1f_24fps.zip](https://github.com/Unity-Technologies/com.unity.formats.alembic/files/6858766/michl_simpleMaterials_1f_24fps.zip) (zipped because Github doesn't want to eat alembics)


Screenshot 2021-07-21 164742

lucillecaillaud:
We have no intention to put more effort into this for now. This PR is linked in the related JIRA [ABC-76](https://jira.unity3d.com/browse/ABC-76) in the comment and can be re-opened when appropriate.

PR #467, opened 2021-08-12
ABC-229 - Document the material remapper tool

sebastienduverne:
[ABC-229](https://jira.unity3d.com/browse/ABC-229) - Update the User Manual to implement info about the new material assignation tool of the Importer:
- [x] Update the Alembic Import Settings window page => Reference
- [x] Update the "Working with Materials" page => Main material-related concept and tasks (also fixes [ABC-212](https://jira.unity3d.com/browse/ABC-212))
- [x] Extract/move "motion blur" and "shaders" from the Materials page => topic re-focus
- [x] Update the "Importing a local Alembic file..." page => Entry point within Import topic for regular case
- [x] Update the Importing an external Alembic file... page => Entry point within Import topic for external case
- [x] Update the Importing Alembic files page => Feature mention in main Import topic entry point
- [x] Update the index page => Feature mention in overall context


com.unity.formats.alembic/Documentation~/ref_Importer.md
```diff
@@ -5,42 +5,82 @@ The Alembic **Import Settings (Alembic Importer)** window allows you to customiz
...
+
+![List of meshes and submeshes with faceset names](images/abc_import_options_materials_meshes_facesets.png)
+
+| Label | Description |
+|:---|:---|
+| (1) | A mesh of the imported Alembic asset.<br /><br />**Note:** The Meshes / Facesets treeview does not represent the actual GameObject hierarchy of the imported Alembic asset. To get hierarchy context for a specific mesh, place your mouse cursor over the mesh name. |
+| (2) | Submeshes of mesh (1), identified by the names of their corresponding Facesets in the original Alembic file.<br /><br />**Note:** "Empty" means the corresponding Faceset has no name in the original Alembic file. In that case, you can only assign a material manually. |
+| (3) | The current material assignation of the submesh.<br /><br />Use the target fields to manually assign materials to the listed submeshes. For automatic name-based material assignation, use [Material Search](#material-search).<br /><br />**Note:** "None (Material)" means the Alembic Importer doesn't override the default behavior, which assigns the Default Material to the submesh. |
```
clusty, line 75: I would call "material assignation" more a "material override".
It might clarify the whole "null means default material" story.

clusty (approved):

com.unity.formats.alembic/Documentation~/ref_Importer.md
cguer: I don't have a strong opinion but I think I agree with Vlad. The value here could potentially be different than the actual material assigned on the mesh.

cguer (approved):
Looks good to me 👍. Although I see that `faceset` is used here instead of `face set` like #456, is this intentional?

PR #492, opened 2022-01-06
JIRA: ABC-283 Fix for 2.3.X: Package fails to compile when removing the analytics module from a project

clusty:
The alembic package implemented analytics, but did not realize that thew analytics was in a separate module of the editor.

This PR adds a dependency on that module:

Steps can be found here: https://jira.unity3d.com/browse/ABC-283

@sebastienduverne Please validate only the changelog.


cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.7f1 (Windows only)

PR #497, opened 2022-01-20
JIRA:ABC-294 Fix for 2.3.X: [Switch]Unity Analytics errors detected when importing Alembic package

clusty:
Apparently there are special requirements for analytics when building on Switch and other consoles.
When switching the build target to Switch for ex, the whole analytics module gets stripped out due to NDA requirements. This happens for all analytics that are collected inside a Runtime assembly.


cguer (approved):
QA approved 👍!

- Tested on 2021.2.9f1 (Windows only)

thomas-tu (approved):
LGTM. Also tested on Windows with Switch target (Unity 2021.2.8), this patch works.

thomas-tu:
For reference: https://docs.unity3d.com/ScriptReference/Analytics.Analytics.html

PR #499, opened 2022-02-16
Add quotations around description field for correct YAML parsing.

cguer:
The Appcollector team notified us that this YAML file wasn't parsing correctly because of missing quotes around the description field. See: https://unity.slack.com/archives/C7KLD3ECA/p1645021912670899


unitysipu (approved):
should work after this 🤞

clusty (approved):

PR #504, opened 2022-06-30
Disable hdrp test in trunk on Mac for release/2.2

windxu88:
**Purpose of this change:**
This is an integration of https://github.com/Unity-Technologies/com.unity.formats.alembic/pull/503 into release/2.2 branch since HDRP test also fails in trunk on Mac there.
The failure is because of the following warning message:
`warning: Assertion failed on expression: 'localPagePos <= page.allocatedSize'
`
We think we can just disable hdrp test in trunk on Mac for now based on:
- The test doesn't test too much. What it does is just:
`yield return new EnterPlayMode(); Assert.IsTrue(true); yield return new ExitPlayMode();
`
- We think the failure is caused by HDRP itself in trunk on Mac.

**JIRA ticket#:** ABC-313 Alembic HDPR tests begin to fail in trunk on Mac (Disable it for now)

**Description of feature/change:**
The test job definition is still there. I just disabled it in the virtual job "test_trigger_hdrp" which is run by nightly trigger.

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -231,6 +231,13 @@ test_trigger_hdrp:
     - .yamato/upm-ci.yml#pack
     {% for editor in editors %}
     {% for platform in platforms %}
+    {%- comment -%}
+    # Disable hdrp test in trunk on Mac for now which fails nightly trigger because: 
+    # First the test doesn't test too much. Second, we believe the failure is caused by HDRP in trunk
+    {%- endcomment -%}
+    {%- if editor.version == 'trunk' and platform.name == 'mac' -%}
+    {%- continue -%}
+    {%- endif -%}
     - .yamato/upm-ci.yml#testProject_hdrp_{{platform.name}}_{{editor.version}}
     {% endfor %}
     {% endfor %}


windxu88:
> From the JIRA task I see that the error is that Bokken returns an OpenGL VM instead of a Metal one. Instead of disabling HDRP tests on macOS, would you mind trying another VM that I confirm returns a Metal API?
> 
> ```
>   - name: mac
>     type: Unity::metal::macmini
>     model: M1
>     image: package-ci/macos-12:v0.0.1-1084810
>     flavor: b1.large
> ```

Good point, I will give a try.

But FYI, before this test began to fail, the error about API OpenGLCore not supported with HDRP is already there, such as [this successful run](https://unity-ci.cds.internal.unity3d.com/job/14275894/artifacts), check "Execution.log".
It is still worth to try a Metal Mac though.

unitybru:
> > From the JIRA task I see that the error is that Bokken returns an OpenGL VM instead of a Metal one. Instead of disabling HDRP tests on macOS, would you mind trying another VM that I confirm returns a Metal API?
> > ```
> >   - name: mac
> >     type: Unity::metal::macmini
> >     model: M1
> >     image: package-ci/macos-12:v0.0.1-1084810
> >     flavor: b1.large
> > ```
> 
> Good point, I will give a try.
> 
> But FYI, before this test began to fail, the error about API OpenGLCore not supported with HDRP is already there, such as [this successful run](https://unity-ci.cds.internal.unity3d.com/job/14275894/artifacts), check "Execution.log". It is still worth to try a Metal Mac though.

We cannot expect an HDRP project in OpenGL to work. https://docs.unity3d.com/Packages/com.unity.render-pipelines.high-definition@14.0/manual/System-Requirements.html
Even if it "passed" previously, we need it to run on a Metal API in order to consider the test results valid.

clusty (approved):
Thanks for the fix.
The HDRP tests don't test that much so I don't think it's a crime to just disable it.

windxu88:
Thanks all for reviewing. I am merging pull/504 into release/2.2 branch now so that the nightly trigger will not fail everyday. I will also create a ticket for the remaining work such as:
- Why there is such a warning `Assertion failed on expression: 'localPagePos <= page.allocatedSize''.` with trunk on Mac
- Update Mac vm to `Unity::metal::macmini` for HDPR tests on Mac

PR #505, opened 2022-07-12
upm-ci can now run tests on centos using only CPU

mihai-unity:
Similar to the change in [upm-ci-yamato-temaplates][1], there's no GPU requirement for running centos tests.

[1]: https://github.cds.internal.unity3d.com/unity/upm-ci-yamato-templates/pull/31


mihai-unity:
@clusty since I don't have write access rights to create branches on the repo, I can't queue any CI to prove the changes are good. Can you do that instead?

clusty:
@mihai-unity Done. Deputized :) 
Should you be able to run yamato from a fork ?

PR #520, opened 2022-12-08
ABC-330 Remove trunk from release 2.2 branch and exclude 2019.4 from clean console test

windxu88:
## Purpose of PR:
Remove trunk from release 2.2 branch because trunk only supports alembic version 2.3.0+.
Also exclude 2019.4 from clean console test because there some logs/warnings in 2019.4, but we don't officially support 2019.4 any more.
Also updated 2020.2 to 2020.3, 2021.1 to 2021.3 in CI.

**JIRA ticket #:**
[ABC-330](https://jira.unity3d.com/browse/ABC-330) Fix Alembic clean console tests

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -1,8 +1,7 @@
 editors:
   - version: 2019.4
-  - version: 2020.2
-  - version: 2021.1
-  - version: trunk
+  - version: 2020.3
+  - version: 2021.3
 
 platforms:
   - name: win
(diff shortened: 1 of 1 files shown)


clusty (approved):

.yamato/upm-ci.yml
```diff
@@ -1,8 +1,7 @@
 editors:
   - version: 2019.4
-  - version: 2020.2
-  - version: 2021.1
-  - version: trunk
+  - version: 2020.3
+  - version: 2021.3
```
clusty, line 4: do we need 2022.2(or 3)

windxu88: I checked [manifest.json of Unity 2022.2 and 2022.1](https://github.cds.internal.unity3d.com/unity/unity/blob/2022.2/staging/External/PackageManager/Editor/manifest.json), its minimally supported version is also 2.3.0. 
It seems 2021.3 is the highest version which still supports `alembic 2.2.2`.

PR #522, opened 2022-12-20
Fix reimport on quality change (ABC-336)

thomas-tu:
JIRA TICKET: https://jira.unity3d.com/browse/ABC-336

Bug:

With HDRP, switching the project quality setting results in a reimport of all Alembic assets.
The reason being that our `AlembicImporter` uses the `currentRenderPipeline` asset to build a dependency hash. In HDRP, by default, each quality level has a specific `RenderPipelineAsset` attached to it so every time we switch between quality level, hash is different. 

This mechanism exists so each imported asset always gets a valid default material.

Solution:
Rely on the default material asset instead of the render pipeline asset. It's unlikely that the default material would change often in the lifetime of a project but in case a team uses their own render pipeline, the imported Alembic will still get a proper material.

Tests:
Manual tests, with HDRP and URP projects.
Steps:
* Switch between Quality level in a project with HDRP.

Doc:
- [x] Changelog
- Does not affect the UI

### DIFF
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [2.3.1] - 2022-12-06
 ### Added
 ### Changed
+- The Editor will stop re-importing Alembic assets when a user changes their project quality settings (HDRP).
 ### Fixed
 
 ## [2.3.0] - 2022-01-28
--- a/com.unity.formats.alembic/CHANGELOG.md
+++ b/com.unity.formats.alembic/CHANGELOG.md
@@ -7,8 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [2.3.1] - 2022-12-06
 ### Added
 ### Changed
-- The Editor will stop re-importing Alembic assets when a user changes their project quality settings (HDRP).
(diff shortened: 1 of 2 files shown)


com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [2.3.1] - 2022-12-06
 ### Added
 ### Changed
+- The Editor will stop re-importing Alembic assets when a user changes their project quality settings (HDRP).
```
clusty, line 10: I think this might be a fixed entry.
Deferring to @MaryPerchanok

clusty (approved):

cguer (approved):
QA approved 👍!

Tested on:
- 2023.1.0a24 (Windows only)

RoseHirigoyen (approved):

com.unity.formats.alembic/CHANGELOG.md
MaryPerchanok: It looks like it reflects the fixed state, but I think it could be phrased better. (change to present tense) 
Suggest"
"When a user changes HDRP quality settings, the Editor no longer re-imports Alembic assets."
(or possibly, "When a user changes HDRP quality settings, the Editor no longer re-imports Alembic assets that already exist in the project.")

MaryPerchanok (approved):

PR #523, opened 2023-01-04
[Documentation] Fix some minor issues in documentation

RoseHirigoyen:
JIRA issues:
- [ABC-258](https://jira.unity3d.com/browse/ABC-258): The Curves component doesn't have a reference page in User Manual
- [ABC-259](https://jira.unity3d.com/browse/ABC-259): The Curves Renderer component doesn't have a reference page in User Manual
- [ABC-260](https://jira.unity3d.com/browse/ABC-260): The Custom Data component doesn't have a reference page in User Manual
- [ABC-261](https://jira.unity3d.com/browse/ABC-261): The Shot Asset screen is wrongly identified as a component in User Manual

What this PR does:
- Added `Alembic Curves`, `Alembic Curves Renderer` and `Alembic Custom Data` options screenshots
- Added `Curves component` reference page
- Added `Curves Renderer component` reference page
- Added `Custom Data component` reference page
- Renamed `Shot Asset component` to `Shot Asset clip settings`
- Added links to new pages in the reference.md table

Also:
- I noticed that not every component that had the `Script` option was referencing it, so I added it where it was missing
- The `Alembic Shot Asset component` description in the reference.md table had a typo (`where where`)


MaryPerchanok (commented):
I'm not sure whether Sebastien mentioned it, but  you should also edit the TOC to add your new topics.

Also, for the Window Reference topic, to comply with our content model, for all of the  right column entries, suggested sentences.

com.unity.formats.alembic/Documentation~/ref_CurvesRenderer.md
```diff
@@ -0,0 +1,11 @@
+# Alembic Curves Renderer component
+
+The Alembic Curves Renderer component allows you to visualize an Alembic curve in a basic way. Adding a Curves Renderer component automatically adds a Mesh Renderer component and a Mesh Filter component.
```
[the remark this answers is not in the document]
RoseHirigoyen: Ooh I thought it was a `single curve` but it is a curve node (so plural)!

RoseHirigoyen:
I noticed that the `Alembic Clip Recorder` is in the TOC but not in `reference.md`, should it be there too? @MaryPerchanok

com.unity.formats.alembic/Documentation~/TableOfContents.md
```diff
@@ -16,10 +16,13 @@
 * [Working with Curves](curves.md)
 * [Applying motion blur and shaders](motion-vectors.md)
 * [Window reference](reference.md)
```
MaryPerchanok, line 18: I wonder whether we can improve the order of the sub topics: group the import and export topics? Not sure whether that makes sense, because I don't know Alembic well enough.

com.unity.formats.alembic/Documentation~/ref_CustomData.md
```diff
@@ -0,0 +1,15 @@
+# Alembic Custom Data component
+
+The Alembic Custom Data component allows you to see additional custom information stored with each Face Set. A Face Set is a sub-group of faces that compose an object.
```
MaryPerchanok, line 3: Suggest: 
he Alembic Custom Data component displays additional custom information stored with each Face Set. A Face Set is a sub-group of faces that compose an object.

com.unity.formats.alembic/Documentation~/TableOfContents.md
RoseHirigoyen: What would you think of something like this (and the table in reference.md could reflect it):

* [Window reference](reference.md)
    * [Alembic Import Settings window](ref_Importer.md)
        * [Alembic Stream Player component](ref_StreamPlayer.md)
    * [Alembic Curves component](ref_Curves.md)
        * [Alembic Curves Renderer component](ref_CurvesRenderer.md)
    * [Alembic Custom Data component](ref_CustomData.md)
    * [Alembic Point Cloud component](ref_PointCloud.md)
        * [Alembic Point Renderer component](ref_PointRenderer.md)
    * [Alembic Exporter component](ref_Exporter.md)
    * [Alembic Clip Recorder](ref_Recorder.md)
        * [Alembic Shot Asset clip settings](ref_Shot.md)

MaryPerchanok: That looks really good! I will check the content model to see if we can have that many indents and get back to you.

MaryPerchanok: Looks like 3 level in the TOC is fine. I'm wondering about where Alembic Shot Asset clip settings belongs. I suspect it's related to Alembic Exporter component, not Alembic Clip Recorder. If it affects both, I would move it to level 2.

RoseHirigoyen: From what I understand, the `Alembic Shot Asset clip settings` is the setting that describes an alembic shot in a Timeline Alembic Track. I would not put them under Exporter, because they are not directly related (unlike the importer and the stream player)

I think we could put the `Alembic Clip Recorder` and the `Alembic Shot Asset clip settings` at level 2, or we could put them at level 3 under a `Timeline` subtitle, for example:
* [Window reference](https://github.com/Unity-Technologies/com.unity.formats.alembic/pull/523/reference.md)
    * ...
    * Timeline
        * [Alembic Clip Recorder](https://github.com/Unity-Technologies/com.unity.formats.alembic/pull/523/ref_Recorder.md)
        * [Alembic Shot Asset clip settings](https://github.com/Unity-Technologies/com.unity.formats.alembic/pull/523/ref_Shot.md)

MaryPerchanok: I would leave Alembic Clip Recorder on level 2 with the Exporter; it exists as a window, where you can use it without a Timeline, and can be added to the Timeline by adding a Recorder track.  The person looking for it would be interested in exporting. (I think I need to update the topic on Alembic Clip Recorder.)
I just reread the existing documentation: the Alembic Shot Asset Clip component is 
 mentioned in the topic "Play back using an Alembic Shot clip", and nowhere else. So if you created a Timeline category, it could go there, but so would Alembic Shot Asset component..

RoseHirigoyen: Sounds good! :)

com.unity.formats.alembic/Documentation~/reference.md
```diff
@@ -2,9 +2,15 @@
 
 This section provides details on the properties for the following windows or components:
 
-- [Alembic Exporter component](ref_Exporter.md), which allows you to export GameObjects, Cameras, and Points (in the Particle System).
-- [Alembic Point Cloud component](ref_PointCloud.md), which allows you to customize the shape and behavior of the point cloud.
-- [Alembic Point Renderer component](ref_PointRenderer.md), which allows you to customize the appearance of the particles in the point cloud.
-- [Alembic Shot Asset component](ref_Shot.md), which allows you to where where the Alembic Shot clip stops playback.
-- [Alembic Stream Player component](ref_StreamPlayer.md), which allows you to play back animation from the Alembic file in either Timeline or Unity Animation.
-- [Import Settings for Alembic](ref_Importer.md), which allows you to customize importing the data from the Alembic file.
+| Element                                                        |                                                                | Description                                                                                                                                          |
```
cguer, line 5: I'm probably missing something obvious here 😅 but what is the difference between the items listed in the first columns vs. the second column? (and if there is a reason, could we add a header for the second column?)

RoseHirigoyen: I was trying to do something like [this file](https://github.com/Unity-Technologies/com.unity.formats.alembic/blob/main/com.unity.formats.alembic/Documentation~/import-options-cameras.md) to show that the components are related 😆 but I can remove the second level if it's clearer!

cguer: Gotcha! I think it makes a bit more sense in that other page because it shows a specific relationship, the left column is the property name and the right column are the possible choices for that property.

I can see the relationship you are representing in this table but I'm not totally sure it's improving the user's understand of how these components are linked. I honestly don't have that strong of an opinion on this 😛  . Either way is totally fine but I would personally lean more towards a single column.

Technically, you can also have an `Alembic Stream Player` without the importer so the relationship is not always there (ex: add a Stream Player component on a gameobject and you can point to an alembic file outside of the project)

RoseHirigoyen: Ahhhh that makes sense! I'll put them back in two columns then :) However, should I place components in the same order as the table of content or in alphabetical order?

cguer: Hmm, good question. Your order makes sense logically but since it's a more "dry" reference page maybe alphabetically is the standard?

@MaryPerchanok do you have an opinion on this? Do you think we should order the components alphabetically or logically (components that are similar or have similar purpose stay together)

sebastienduverne: My two cents here: consider the reference section like a dictionary, so its TOC structure should be in alphabetical order. Other perspectives that are more contextual/conceptual (similar purpose or complementary roles) would be addressed in specific topics of the doc.

cguer (approved):
QA approved 👍! Thanks for jumping in on this

MaryPerchanok (approved):

PR #531, opened 2023-02-28
ABC-374: Update bokken images for Alembic CI

windxu88:
## Purpose of this PR:
Currently bokken images in Alembic CI are deprecated. Need to replace them with new ones.
Also trunk has crash issues with running in player mode on mac:stable (macos 10.15), because of this, we also need to update Mac image to macos-12.

In this PR, I also added `set -e` in build.sh so that the build script won't fail silently.

**JIRA ticket:**
[ABC-374](https://jira.unity3d.com/browse/ABC-374) CI: Update bokken images for Alembic CI

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -8,13 +8,11 @@ editors:
 platforms:
   - name: win
     type: Unity::VM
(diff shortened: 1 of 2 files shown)


RoseHirigoyen (approved):

lucillecaillaud (approved):
Looks good, thanks Wind!

PR #534, opened 2023-03-17
ISPC - Clamp01 returned values from dot() to avoid passing invalid values to acos() (ABC-383)

thomas-tu:
Jira: https://jira.unity3d.com/browse/ABC-383

Our implementation of angle_between on the ISPC layer can sometimes return a floating point that falls outside of the range of [-1; 1]. Then, when we pass such value to `acos` - which expects a value within the range [-1; 1] - it returns NaN.

Example: 1.0000001f is enough to get a NaN.

ISPC angle_between implementation should now be on par with the [generic implementation](https://github.com/Unity-Technologies/com.unity.formats.alembic/blob/751a7260cc4247c322dee0434956426759fda9e3/Source/abci/Foundation/aiMath.h#L529), which has no issue for the case we are fixing.

This issue has been brought up by Solutions. They've also validated that this fix work for them.


RoseHirigoyen (approved):
LGTM!

lucillecaillaud (approved):

PR #539, opened 2023-04-04
ABC-386 Set MACOSX_DEPLOYMENT_TARGET to 10.14

windxu88:
## Purpose of this PR:
In Alembic CI, mac image has been update to macos-12. But that will cause problems in `build on Mac` job because the result libs will not be able to load on Mac 10.14+ and Mac 11 which Unity still supports.
Set MACOSX_DEPLOYMENT_TARGET to 10.14.

I am using environment variable `MACOSX_DEPLOYMENT_TARGET`, because:
- If `CMAKE_OSX_DEPLOYMENT_TARGET` is not set explicitly, it will be initialized with `MACOSX_DEPLOYMENT_TARGET`.
- Together with building alembic libs, other dependency libs such as openexr also needs to be built. Using `MACOSX_DEPLOYMENT_TARGET` is the easiest way for all build settings. It will apply to all libs.

I have tested the libs built on Mac 12 with this environment variable `MACOSX_DEPLOYMENT_TARGET` can be loaded on Mac 10.15.

**JIRA ticket:**
[ABC-386](https://jira.unity3d.com/browse/ABC-386) CI: Libs built on macos-12 cannot be loaded on mac 11, need to set CMAKE_OSX_DEPLOYMENT_TARGET to 10.14

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,11 +9,11 @@ set -e
 export CXXFLAGS="-O3 -fomit-frame-pointer -fPIC"
 export CFLAGS="-O3 -fomit-frame-pointer -fPIC"
 
-OSX_DEPLOYMENT_TARGET=
+OSX_DEPL_TARGET=
 if [ "$(uname)" == "Darwin" ]; then
    export CXXFLAGS="${CXXFLAGS} -arch x86_64 -arch arm64"
    export CFLAGS="${CFLAGS} -arch x86_64 -arch arm64"
-   OSX_DEPLOYMENT_TARGET=10.14
+   OSX_DEPL_TARGET=10.14
 
 fi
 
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -12,6 +12,7 @@ export CFLAGS="-O3 -fomit-frame-pointer -fPIC"
 if [ "$(uname)" == "Darwin" ]; then
    export CXXFLAGS="${CXXFLAGS} -arch x86_64 -arch arm64"
    export CFLAGS="${CFLAGS} -arch x86_64 -arch arm64"
+   export MACOSX_DEPLOYMENT_TARGET=10.14
```
lucillecaillaud, line 15: Looks good to me. But @RoseHirigoyen also told me that someone recommended her to actually build FROM a machine that has the right version if possible.

I personally don't have enough knowledge to really know the differences between the two.

lucillecaillaud (approved):

build.sh
RoseHirigoyen: Or you can use bee to download the target version SDK - I didn't do it for recorder because it is a lot of setup and the `MACOSX_DEPLOYMENT_TARGET` ended up working well, but it's a thing to keep in mind if we have unexplainable dllnotfound errors in the future!

RoseHirigoyen (approved):

PR #541, opened 2023-05-18
ABC-333 CI: Refactor Alembic's CI

windxu88:
## Purpose of this PR:
Refactor Alembic CI so jobs of similar functions will be grouped together in separate files.

**JIRA ticket:**
[ABC-333](https://jira.unity3d.com/browse/ABC-333) CI: Refactor Alembic's CI


.yamato/package-pack.yml
```diff
@@ -0,0 +1,24 @@
...
+     paths:
+       - "upm-ci~/packages/**/*"
+   build:
+     paths:
+       - "com.unity.formats.alembic/**/*"
+  dependencies:
+    - .yamato/build-plugins.yml#build_win
+    - .yamato/build-plugins.yml#build_mac
+    - .yamato/build-plugins.yml#build_centOS
+
```
RoseHirigoyen, line 24: Nit: newline
(I think line 24 has spaces maybe? or if git is doing something weird ignore my comment)

RoseHirigoyen (approved):
LGTM!

.yamato/package-pack.yml
windxu88: You are right, it is two spaces. Remove them.

.yamato/global.metafile
```diff
@@ -0,0 +1,83 @@
+test_editors:
+  - version: 2019.4
+  - version: 2020.3
```
thomas-tu, line 3: We can drop 2019.4. End of life of 2020.3 is also coming soon.

thomas-tu (approved):

.yamato/global.metafile
windxu88: In Alembic's package.json, the minimum Unity version is still `"unity": "2019.4",`, because of that we cannot remove 2019.4 from CI, otherwise publishing workflow will fail.

thomas-tu: Good to know, thanks.

PR #547, opened 2023-06-12
ABC-389 Add "Generate documentation" job to Alembic package

windxu88:
## Purpose of this PR:
Add "Generate documentation" job to Alembic package

**JIRA ticket:**
[ABC-389](https://jira.unity3d.com/browse/ABC-389) CI: Add "Generate documentation" to Alembic package

### DIFF
--- a/.yamato/package-documentation.yml
+++ b/.yamato/package-documentation.yml
@@ -1,0 +1,17 @@
+{% metadata_file .yamato/global.metafile %}
+---
+
+generate_documentation:
+  name : Generate documentation
+  agent:
+    type: {{ mac_platform.type }}
+    image: {{ mac_platform.image }}
+    flavor: {{ mac_platform.flavor }}
+  commands:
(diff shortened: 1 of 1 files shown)


windxu88:
The generated doc is in [the artifacts of Generate documentation job](https://unity-ci.cds.internal.unity3d.com/job/25587564/artifacts).

thomas-tu (approved):

thomas-tu:
LGTM but does it have to be on Mac? Can't we run on it on a more accessible OS?

windxu88:
> LGTM but does it have to be on Mac? Can't we run on it on a more accessible OS?
Contrary to what many people think, Mac VM is actually the platform with least distribution time. I mean pure Mac VM, not Mac VM plus a mac mini device. Usually Mac VM's distribution time is only several minutes VS Win and Ubuntu's tens of minutes.
I think that's because most Yamato jobs are running on Window and Ubuntu platforms.

That's also why I choose Mac as the first platform the brick supports:). Window and Linux support is on my to-do list.

PR #548, opened 2023-06-13
Fix broken reference when an Alembic file is moved (ABC-370)

thomas-tu:
## Purpose of this PR

**Ticket/Jira #:** https://jira.unity3d.com/browse/ABC-370

This PR makes sure that `AlembicStreamDescription.PathToAbc` is properly updated when an Alembic file is moved on the filesystem. We set a dependency to the asset path. As soon as it changes, the asset is reimported. There's probably ways to handle this more gracefully but that might take some adjustment on our data model.

## Testing

**Functional Testing status:** Added an automatic test to cover this case.

**Performance Testing status:** N/A

## Overall Product Risks

**Complexity:** Minimal

**Halo Effect:** Minimal

## Documentation & UX Writing

| User facing text to review |  | Details |
| :--- | :--- | :--- |
| User interface | N |  |
| Public API docs | N |  |
| Changelog | Y |  |

| Documentation halo effect |  | Jira link |
| :--- | :--- | :--- |
| User manual | N |  |
| Other docs | N |  |


com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Prevent a warning in the Console due to deprecated `RecorderInputSetting.ValidityCheck()` when Alembic is installed along with Recorder.
 - Prevent warnings related to FindObjectOfType<> calls on package installation with Unity Editor 2023.1+.
+- Prevent broken links in Alembic files that had been imported then moved outside of the Unity Editor.
```
sebastienduverne, line 11:
```suggestion
- Prevent broken links in imported Alembic files that have been moved outside of the Unity project folder.
```

sebastienduverne (commented):
Changelog suggestion

com.unity.formats.alembic/CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Prevent a warning in the Console due to deprecated `RecorderInputSetting.ValidityCheck()` when Alembic is installed along with Recorder.
 - Prevent warnings related to FindObjectOfType<> calls on package installation with Unity Editor 2023.1+.
-- Prevent broken links in Alembic files that had been imported then moved outside of the Unity Editor.
+- Prevent broken links in imported Alembic files that have been moved outside of the Unity project folder.
```
sebastienduverne, line 11:
```suggestion
- Prevent broken links in Alembic files that would be first imported and then moved outside of the Unity project folder.
```

sebastienduverne (commented):
Second take for clarification, hope it helps :)

cguer (approved):
QA approved 👍!

Tested on:
- 2020.3.48f1 (Windows only)
- 2021.3.27f1 (Windows only)
- 2022.3.2f1 (Windows only)
- 2023.2.0a18 (Windows only)

cguer (changes requested):
Sorry, I missed something in my review 😅. It looks like there is an issue when closing/reopening the editor.

Reproduction steps (also in attached video):

1. Import an alembic file into your project
2. Open or create a scene
3. Instantiate the alembic file into the current scene
4. In a file browser, navigate to the project's directory where the alembic file is
5. From the file browser, move the alembic file _and_ its .meta file into a different directory (but still under /Assets)
6. Close the Unity Editor
7. Re-open the Unity Editor

https://github.com/Unity-Technologies/com.unity.formats.alembic/assets/3811928/bb28676c-ed2b-4424-bec6-150c81a7154f

com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
```diff
@@ -20,6 +19,22 @@
 [assembly: InternalsVisibleTo("Unity.Formats.Alembic.UnitTests.Editor")]
 namespace UnityEditor.Formats.Alembic.Importer
 {
+    class AlembicAssetPostProcessor : AssetPostprocessor
+    {
+#if UNITY_2021_3_OR_NEWER
+        static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths, bool didDomainReload)
+#else
+        static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths)
+#endif
```
lucillecaillaud, line 28: Are we really supporting version older than 2021.3??

Technically, Alembic 2.3.X is only released on 2022.3+. Should we simply update the minimum version allowed and don't bother with `#if/#else`?

com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
```diff
@@ -20,6 +19,22 @@
...
+        {
+            foreach (var path in movedAssets)
+            {
+                if (path.EndsWith(".abc", StringComparison.OrdinalIgnoreCase))
+                    AssetDatabase.ImportAsset(path);
```
lucillecaillaud, line 33: Could it be a bug in the asset database? It feels strange to have to do that ourselves.
Is the Alembic importer that odd? Or is it a special case?

thomas-tu: It's a specific from our Alembic plugin. We store a piece of data in one of the Alembic SubAsset that needs to be updated if the Alembic file is moved. The ADB is doing its job as expected. We're actually missing a proper way to handle this case.

I'm trying to figure if there's a better way to handle this but at the current state of the fix, I would simply trigger a reimport of the asset. Without this PR, the fix/workaround for users is to also `Right click > Reimport` on the affected assets.

com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
lucillecaillaud: It is supporting it, but it's not released in it. So by default, when a user open Unity 2020.3, it's not the version they'll have.
But yes, I forgot that bumping the Unity version requires a new major. So I guess we have to stick to #if/#else for now 😞

com.unity.formats.alembic/Editor/Importer/AlembicImporter.cs
lucillecaillaud: I thought sub-assets would be re-imported as well as the main asset when you move it. Actually, if the main asset is correctly re-imported when it's moved, it should "re-build" the sub-assets no? As they're part of the importer (in a normal case at least).

If the sub-asset is not generated by the importer, but later... can't there be at least a dependency between the two? (I'm not entirely sure how that work, maybe it doesn't make sense, I just know you can set dependencies between imported assets so that when one is re-imported the other is as well).

(we can do a quick huddle if it's simpler, I'm curious to understand the case so that I also better understand how the asset database works)

thomas-tu: I pushed an update in [a8c902c](https://github.com/Unity-Technologies/com.unity.formats.alembic/pull/548/commits/a8c902c5f1bf5e96a926b148ec18109b058381a2).
It sets a dependency with the assetPath more nicely. No more tricks!

cguer (approved):
QA approved 👍!

Tested on:
- 2023.2.0b4 (Windows only)

lucillecaillaud (approved):
Didn't test, but code looks good to me!

RoseHirigoyen (approved):
LGTM!

PR #549, opened 2023-06-14
ABC-403 Exclude 2019.4 from HDRP tests

windxu88:
## Purpose of this PR:
We are trying to upgrade AlembicHDRP project to 22.2, but it will be compatible with 2019. Since 2019 is too old and unsupported, we just decide to exclude it from HDRP test.

**JIRA tickets:**
[ABC-403](https://jira.unity3d.com/browse/ABC-403) Upgrade AlembicHDRP project to 22.2

### DIFF
--- a/.yamato/package-triggers.yml
+++ b/.yamato/package-triggers.yml
@@ -49,7 +49,10 @@ nightly_test_trigger:
 {% endfor %}
 
 {% for editor in test_editors %}
+# Exclude 2019.4 from HDRP tests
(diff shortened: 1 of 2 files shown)


windxu88:
@thomas-tu , I have created this PR to remove 2019.4 from HDRP tests. Please review. After that, I think you can go ahead with updating HDRP test project. Thanks

thomas-tu (approved):
Thanks Wind!

PR #584, opened 2024-01-13
[ABC484] Handle polymesh interpolation 

SanaJal:
## Purpose of this PR

**Ticket/Jira #: [ABC-484](https://jira.unity3d.com/browse/ABC-484)**
This PR handles the interpolation of polymesh. 

## Testing

**Functional Testing status:**
Manual testing using the following asset : 
[test_asset_interpolation.zip](https://github.com/Unity-Technologies/com.unity.formats.alembic/files/13925288/test_asset_interpolation.zip)

This gives the following result : 
interpolation


thomas-tu (approved):
Could you comment out the behaviour for interpolating matrices? as discussed I'm not sure this is the right way to do it.

PR #604, opened 2024-09-12
Various minor documentation fixes

sebastienduverne:
## Purpose of this PR

Fix various minor and old issues in Alembic package docs:

- [ ] [DOCATT-5001](https://jira.unity3d.com/browse/DOCATT-5001) - Document the behavior of Alembic SubD import
- [ ] [DOCATT-5004](https://jira.unity3d.com/browse/DOCATT-5004) - Add Render Frame Step to Alembic Clip Recorder properties
- [ ] [DOCATT-5011](https://jira.unity3d.com/browse/DOCATT-5011) - User feedback/audit: fix the "Using Alembic with Unity Animation" page

### DIFF
--- a/com.unity.formats.alembic/Documentation~/import-options-scene.md
+++ b/com.unity.formats.alembic/Documentation~/import-options-scene.md
@@ -10,7 +10,7 @@ Use these options to customize the scale factor, handedness, interpolation, and 
 | __Interpolate Samples__ | Enable to interpolate animation for Transform, Camera, and Mesh components for which the topology does not change (that is, for which the number of vertices and indices are immutable).<br/>If **Interpolate Samples** is enabled, or velocity data is included in the .abc file, you can pass velocity data to an Alembic shader. |
 | __Import Visibility__ | Enable to import Visibility (i.e. the control of the active state of the objects). |
 | __Import Camera__ | Enable to import Camera data. |
-| __Import Meshes__ | Enable to import Mesh data. |
+| __Import Meshes__ | Enable to import Mesh data.<br />**Note:** If a mesh includes SubD data, the importer discards this data and converts the base mesh to make it compatible with the Unity Editor. |
 | __Import Points__ | Enable to import Point (particle cloud) data. |
 | __Import Curves__ | Enable to import Curve (hair or fur) data. See [Working with Curves](curves.md) for more details. |
 | __Add Curve Renderers__ | Enable to get a [basic preview](curves.md#rendering-alembic-curves) of the Alembic curves in your Scene and Game views if you don't have a curve renderer.<br />This option is only available when you select __Import Curves__. |


com.unity.formats.alembic/Documentation~/animClip.md
```diff
@@ -3,12 +3,16 @@
 You can control the playback of Alembic using Unity's Animation System bound to the GameObject that contains the [Alembic Stream Player](ref_StreamPlayer.md) component.
 
 1. Select the GameObject with the Alembic Stream Player component and open the [Animation window](https://docs.unity3d.com/Manual/animeditor-UsingAnimationEditor.html).
-2. [Create an Animator component and an Animation clip](https://docs.unity3d.com/Manual/animeditor-CreatingANewAnimationClip.html) if you don't already have them.
-3. [Save keyframes](https://docs.unity3d.com/Manual/animeditor-AnimatingAGameObject.html) for the **Time** property on the [**Alembic Stream Player**](ref_StreamPlayer.md) component.
+
+1. [Create an Animator component and an Animation clip](https://docs.unity3d.com/Manual/animeditor-CreatingANewAnimationClip.html) if you don't already have them.
```
mfe, line 7: A little weird that every item now is suffixed "1.". 
But the rendered version is good, though.

com.unity.formats.alembic/Documentation~/import-options-scene.md
```diff
@@ -10,7 +10,7 @@ Use these options to customize the scale factor, handedness, interpolation, and
 | __Interpolate Samples__ | Enable to interpolate animation for Transform, Camera, and Mesh components for which the topology does not change (that is, for which the number of vertices and indices are immutable).<br/>If **Interpolate Samples** is enabled, or velocity data is included in the .abc file, you can pass velocity data to an Alembic shader. |
 | __Import Visibility__ | Enable to import Visibility (i.e. the control of the active state of the objects). |
 | __Import Camera__ | Enable to import Camera data. |
-| __Import Meshes__ | Enable to import Mesh data. |
+| __Import Meshes__ | Enable to import Mesh data.<br />**Note:** If a mesh includes SubD data, the importer discards this data and converts the base mesh to make it compatible with the Unity Editor. |
```
mfe, line 13: My understanding is that the SubD data is converted into something compatible with the editor.
But I'm a little confused by this sentence, when I read it, it feels the data is not used at all and I'm not sure it is clear what is converted.
It would probably clearer to say the data is discarded after we say we do a conversion.

com.unity.formats.alembic/Documentation~/animClip.md
sebastienduverne: Actually, this is the way we should always identify a numbered line in markdown. It's easier to maintain the source (insertion, removal, reordering) and the rendered version takes care of the actual numbering.

com.unity.formats.alembic/Documentation~/import-options-scene.md
sebastienduverne: You're right, after re-reading, it feels weird, I should reword.
The conclusion of investigations with Thomas (plus some research I made about SubD):
- SubD involves data and process to calculate smoother surfaces (inner mesh) between vertices/edges of a base polygonal mesh (outer mesh).
- Before, the importer was simply not supporting Alembic files containing any SubD data. No possible import, because Unity doesn't support SubD.
- Now, such files can be imported but Unity keeps only the outer mesh and discards anything normally used to calculate the inner mesh.

mfe (approved):
LGTM

PR #609, opened 2025-02-04
DOCATT-5002 - Improve Alembic Recorder topic

sebastienduverne:
## Purpose of this PR

[DOCATT-5002](https://jira.unity3d.com/browse/DOCATT-5002) - Improve Alembic Clip Recorder topic
* The Recorder window screenshot was already up to date.
* Split the existing task in two (Recorder window vs. Recorder clip), aligning with what was done in the FBX Exporter docs for the same topic.
* Adjusted minimal content to the Alembic context (using the existing – now former – task), including the Recorder clip screenshot.
* Converted the initial file to a landing page.

## Note to reviewers

We're getting rid of old but straightforward documentation tickets that still make sense in KTLO context. This PR doesn't include actual new content and is only about restructuring and alignment.


com.unity.formats.alembic/Documentation~/ref_Recorder.md
mfe, line 1: Line 16 there is a `(The [Input](#input) section is only relevant when you set the recording section's Playback property to Variable, and the Output Format section is always empty for this type of recorder.)`

Feels like this should be a note and that we might want to explain why for Output Format section is empty.

mfe (commented):
Minor comment on a section that was not modified.
Except that LGTM

mfe (approved):

PR #627, opened 2026-03-24
 [UUM-137643] Fix compilation errors and warnings

mfe:
## Purpose of this PR

**Ticket/Jira #:** [UUM-137643](https://jira.unity3d.com/browse/UUM-137643)

This PR address 3 issues:

### 1 - Compilation errors in the console
`error CS0619: 'Object.GetInstanceID()' is obsolete: 'GetInstanceID is deprecated. Use GetEntityId instead.`

### 2 - Warnings (that will become errors in a foreseeable futur):
`warning CS0618: 'FindObjectsSortMode' is obsolete: 'FindObjectsSortMode has been deprecated. Use the FindObjectsByType overloads that do not take a FindObjectsSortMode parameter.' 
`
⚠️ In previous versions of the code, a sorting by FindObjectsSortMode.instanceID was used to have the determinism needed for UpdateCaptureNodes. But that was an inconsistant way of sorting. Furthermore, with entityID, the small amount of creation-order semantics that were accidentally present in instanceID is removed. That's why this PR instroduces another way of sorting with determinism (cf `SortComponentsByStableSceneHierarchy`).

### 3- replace renderPipelineAsset by defaultRenderPipeline
defautRenderPipeline was introduced in 2019 stream so that was due.

A general mindset for this PR is the less we touch, the better. 
The reason is that this code is very legacy and in basic support. We want to mitigate the risk of breaking the past versions and increase our support load. 
That's why every changes needed were encapsulated in `#if UNITY_6000_4_OR_NEWER...#endif` even when the changes would be considered a better practice for precedent versions too (like not sorting with instanceIds). 

## Testing

**Functional Testing status:**

**Performance Testing status:**

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Documentation & UX Writing

| User facing text to review |  | Details |
| :--- | :--- | :--- |
| User interface |  |  |
| Public API docs |  |  |
| Changelog |  |  |

| Documentation halo effect |  | Jira link |
| :--- | :--- | :--- |
| User manual |  |  |
| Other docs |  |  |

## Additional information

**Note to reviewers:**

**Reminder:**

- [x] Add entry in CHANGELOG.md 
- [x] check defaultRenderPipeline availability - Was introduced as early as 2019.4 so before our minimal supported version (21.3)
- [x] Wait for Recorder 5.1.6 to be released (see [UUM-137641](https://jira.unity3d.com/browse/UUM-137641) - Cf Recorder tests
- [x] Sanity check QA - test import + export both for local and streamed assets with depth

### DIFF
--- a/.yamato/global.metafile
+++ b/.yamato/global.metafile
@@ -1,24 +1,24 @@
 test_editors:
   - version: 2021.3
-  - version: 2022.3
   - version: 6000.0
-  - version: 6000.2
   - version: 6000.3
   - version: 6000.4
+  - version: 6000.5
   - version: trunk
 
 nightly_test_editors:
-  - version: 2022.3
   - version: 6000.0
-  - version: 6000.2
   - version: 6000.3
   - version: 6000.4
-  - version: trunk  
+  - version: 6000.5
+  - version: trunk
 
 clean_console_test_editors:
   - version: 6000.0
-  - version: 6000.2
-  - version: trunk  
+  - version: 6000.3
+  - version: 6000.4
+  - version: 6000.5
+  - version: trunk
 
 promotion_test_editors:
   - version: 2021.3
--- a/com.unity.formats.alembic/Runtime/Scripts/Analytics/AlembicExporterAnalytics.cs
+++ b/com.unity.formats.alembic/Runtime/Scripts/Analytics/AlembicExporterAnalytics.cs
@@ -43,7 +43,9 @@ namespace UnityEngine.Formats.Alembic.Exporter
             {
                 return settings.TargetBranch != null ? settings.TargetBranch.GetComponentsInChildren<T>() : Enumerable.Empty<T>();
             }
-#if UNITY_2023_1_OR_NEWER
+#if UNITY_6000_4_OR_NEWER
+            return Object.FindObjectsByType<T>();
+#elif UNITY_2023_1_OR_NEWER
             return Object.FindObjectsByType<T>(FindObjectsSortMode.InstanceID);
 #else
             return Object.FindObjectsOfType<T>();
--- a/com.unity.formats.alembic/Runtime/Scripts/Analytics/AlembicExporterAnalytics.cs
+++ b/com.unity.formats.alembic/Runtime/Scripts/Analytics/AlembicExporterAnalytics.cs
@@ -44,6 +44,7 @@ namespace UnityEngine.Formats.Alembic.Exporter
                 return settings.TargetBranch != null ? settings.TargetBranch.GetComponentsInChildren<T>() : Enumerable.Empty<T>();
             }
 #if UNITY_6000_4_OR_NEWER
+            // Analytics only counts component instances; order does not affect the result.
             return Object.FindObjectsByType<T>();
 #elif UNITY_2023_1_OR_NEWER
             return Object.FindObjectsByType<T>(FindObjectsSortMode.InstanceID);
--- a/com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
+++ b/com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
@@ -953,7 +953,9 @@ namespace UnityEngine.Formats.Alembic.Util
             if (m_settings.Scope == ExportScope.TargetBranch && TargetBranch != null)
                 return TargetBranch.GetComponentsInChildren(type);
             else
-#if UNITY_2023_1_OR_NEWER
+#if UNITY_6000_4_OR_NEWER
+                return Array.ConvertAll<UnityEngine.Object, Component>(GameObject.FindObjectsByType(type), e => (Component)e);
+#elif UNITY_2023_1_OR_NEWER
                 return Array.ConvertAll<UnityEngine.Object, Component>(GameObject.FindObjectsByType(type, FindObjectsSortMode.InstanceID), e => (Component)e);
 #else
                 return Array.ConvertAll<UnityEngine.Object, Component>(GameObject.FindObjectsOfType(type), e => (Component)e);
--- a/com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
+++ b/com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
@@ -548,7 +548,7 @@ namespace UnityEngine.Formats.Alembic.Util
                     return;
                 }
 #if UNITY_6000_4_OR_NEWER
-                abcObject = parent.abcObject.NewXform(target.name + " (" + EntityId.ToULong(target.GetEntityId()).ToString("X8") + ")", timeSamplingIndex);
+                abcObject = parent.abcObject.NewXform(target.name + " (" + EntityId.ToULong(target.GetEntityId()).ToString("X16") + ")", timeSamplingIndex);
 #else
                 abcObject = parent.abcObject.NewXform(target.name + " (" + target.GetInstanceID().ToString("X8") + ")", timeSamplingIndex);
 #endif
@@ -965,7 +965,7 @@ namespace UnityEngine.Formats.Alembic.Util
 #if UNITY_6000_4_OR_NEWER
         /// <summary>
         /// Sorts <paramref name="components"/> in place into a stable order derived from each object’s scene and
-        /// transform hierarchy (see <see cref="AppendStableHierarchySortKey"/>).
+        /// transform hierarchy (see <see cref="AppendStableHierarchySortKey(StringBuilder, Component, List{int})"/>).
         /// </summary>
         /// <remarks>
         /// <para>
@@ -976,8 +976,9 @@ namespace UnityEngine.Formats.Alembic.Util
         /// </para>
         /// <para>
         /// Implementation: allocate a parallel <c>string[]</c> of the same length, fill each entry by clearing a shared
-        /// <see cref="StringBuilder"/>, calling <see cref="AppendStableHierarchySortKey(StringBuilder, Component)"/>, and
-        /// assigning <c>keys[i] = sb.ToString()</c>. Then <c>Array.Sort(keys, components, StringComparer.Ordinal)</c>
+        /// <see cref="StringBuilder"/> and a shared <c>List&lt;int&gt;</c> scratchpad, calling
+        /// <see cref="AppendStableHierarchySortKey(StringBuilder, Component, List{int})"/>, and assigning
+        /// <c>keys[i] = sb.ToString()</c>. Then <c>Array.Sort(keys, components, StringComparer.Ordinal)</c>
         /// reorders <paramref name="components"/> to match ascending key order. Arrays of length 0 or 1, or a null
         /// reference, are left unchanged without allocating.
         /// </para>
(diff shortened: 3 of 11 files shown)


.yamato/global.metafile
```diff
@@ -1,24 +1,24 @@
 test_editors:
   - version: 2021.3
```
LeoUnity, line 2: shouldn't we also remove this and upgrade the minimum unity version to 6000.0?

Same with romotion_test_editors

LeoUnity (approved):
Changes looks good to me, we do have a test failure on win trunk unfortunately.

.yamato/global.metafile
mfe: It's the minimum version of the package, so we have to keep the job for promotion.

com.unity.formats.alembic/Runtime/Scripts/Analytics/AlembicExporterAnalytics.cs
```diff
@@ -43,7 +43,9 @@ static IEnumerable<T> GetTargets<T>(this AlembicRecorderSettings settings) where
             {
                 return settings.TargetBranch != null ? settings.TargetBranch.GetComponentsInChildren<T>() : Enumerable.Empty<T>();
             }
-#if UNITY_2023_1_OR_NEWER
+#if UNITY_6000_4_OR_NEWER
+            return Object.FindObjectsByType<T>();
```
mfe, line 48: The `FindObjectsSortMode.InstanceID` used in `UNITY_2023_1_OR_NEWER` code path is not needed as the result of `GetTargets` is only used with Any(), so I just removed the sorting mode here.

I was tempted to remove it in `UNITY_2023_1_OR_NEWER`  code path too but on a very legacy package in basic support, the less we touch, the better.
Feel free to push back.

com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
```diff
@@ -947,16 +962,65 @@ public static void ForceDisableBatching()
 
 #endif
 
+#if UNITY_6000_4_OR_NEWER
+        /// <summary>
+        /// Deterministic order for export: loaded scene (path, build index), then hierarchy (sibling indices from root).
+        /// Does not use InstanceID/EntityId ordering (see Unity 6+ engine guidance).
+        /// </summary>
+        static void SortComponentsByStableSceneHierarchy(Component[] components)
```
mfe: Sorting by entityId is discouraged but determinism is needed for `UpdateCaptureNodes`. So introducing an alternative way to sort.

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
```diff
@@ -222,7 +232,11 @@ void ClearMotionVectors(AlembicTreeNode node)
         public bool AbcLoad(bool createMissingNodes, bool serializeMesh)
         {
             m_time = 0.0f;
+#if UNITY_6000_4_OR_NEWER
+            m_context = new SafeContext(aiContext.Create(AllocateNativeContextUid()));
```
mfe, line 239: The native function called at the end of the line is `aiContextCreate(int uid)`. So if we keep using a ulong here, we ultimately have to cast it, which can lead to some narrowing and rare but potential collisions. 
But in the code where the id is used, there is no need for a correlation between the id and the abcTreeRoot object, we just need an id to destroy on dispose (cf `aiContextDestroy(aiContext* ctx)`).

Thus the introduction of an alternative surrogate id.

A question for reviewers: what is the bigger risk ?
1- Changing this code
2- Narrowing and collision
😅

com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
```diff
@@ -947,16 +962,97 @@ public static void ForceDisableBatching()
 
 #endif
 
+#if UNITY_6000_4_OR_NEWER
```
mfe, line 965: Explaination about why we need this:
⚠️ In previous versions of the code, a sorting by FindObjectsSortMode.instanceID was used to have the determinism needed for UpdateCaptureNodes. But that was an inconsistant way of sorting. Furthermore, with entityID, the small amount of creation-order semantics that was accidentally present in instanceID is removed. That's why this PR instroduces another way of sorting with determinism,

.yamato/global.metafile
LeoUnity: I believe its allow to bump the minimum version of the package if said version is not supported anymore without being considered a new major release. But if that would increase the scope of this PR, we can do it separately

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
```diff
@@ -87,6 +90,16 @@ public void Execute()
...
         static List<AlembicStream> s_streams = new List<AlembicStream>();
 
+#if UNITY_6000_4_OR_NEWER
+        // Native aiContext keys are int; use a monotonic surrogate instead of narrowing EntityId (ulong → int
+        // would silently discard the upper 32 bits). Starts at 0; Interlocked.Increment returns 1 on the
+        // first call, so UID 0 is never issued — matching GetInstanceID's guarantee for live objects.
+        // Theoretical int overflow after ~2 billion AbcLoad calls is impossible in practice.
+        static int s_nextNativeContextUid;
+
+        static int AllocateNativeContextUid() => Interlocked.Increment(ref s_nextNativeContextUid);
```
LeoUnity, line 100: Should/could this be uint? we are leaving half of its range on the table

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
LeoUnity: I think this approach you took is great!

com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
LeoUnity: A concern that jumped out to me is performance.
When and how often is this sorting taking place? We are doing a lot more computations including string operations to get this determinism.

Also out of curiosity why is determinism important here? What would it mean not having determinism?

com.unity.formats.alembic/Runtime/Scripts/Importer/AlembicStream.cs
mfe: I'm using the type of the native function that this id is for (aka `aiContextCreate(int uid)`)

.yamato/global.metafile
mfe: I would say it's  indeed out of scope, so let's log a task to update this separately.

com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
mfe: Good questions :D

Put on your seat belt 😄

**Some context**

GetTargets (where this func is used) is the scene discovery mechanism. Its results feed ConstructTree, which inserts nodes into m_nodes aka builds the in-memory capture tree and ultimately determines the order in which Alembic nodes (NewXform, NewPolyMesh, etc.) are created inside the archive.

GetTargets is called at every capture frame during recording.

**Why we want determinism**

The tree topology is naturaly done and deterministic for parents/children because of the recursiveness used (aka we always resolves parents before children).

But what is not naturally stable is the sibling ordering, aka the order in which NewXform/NewPolyMesh is called on the children of a given parent, which determines the order those children appear in the Alembic file. If no sorting is done, sibling order will not be stable and produce .abc files that bit-wise/ascii-wise looks different when they technically are not (noise during versioning, unnessary cache invalidation...).

BTW the sorting by InstanceID was actually not great either, because from a Unity session to another, the instanceID could change and so the ordering of the sibblings.

The new sorting is based on the GO scene path and root-to-leaf sibling indices (that order is serialized into the scene file and restored identically every time the scene loads), this is now a structural and session-independent sorting. 

**Perf concern**

At every captured frame, we unconditionally calls UpdateCaptureNodes, which calls GetTargets once per enabled capturer type (up to 4: Transform, Camera, MeshRenderer, SkinnedMeshRenderer) = SortComponentsByStableSceneHierarchy runs up to 4× per frame for the entire duration of recording 😅

So, ya, not great, the new sorting is probably worse than native InstanceID sorting.
That said there is room for improv, let me try.

mfe: Well, I tried for few hours but the improvements I could get (without breaking everything) are almost anecdoctical and there is still this idea of minimalizing impact / halo in the changes we do in this package.

To mitigate, this is only run when we record. So I would stop there.
Feel free to push back if you feel I should persevere in trying to improve the perf.

How about that @LeoUnity ?

LeoUnity: Seat Belts on 🎽 😁

That makes total sense, thanks for bringing me up to speed on this!

Let's just confirm expectations users might have around this package, and more specifically when this code is running to make sure we can pick an appropriate path forward.
 I see this code lives in the Runtime folder, is there any expectations that this code is run in production with real-time "guarantees"? Or is the expectations around exporting an Alembic file to be done in a non-interactive non-realtime context?

If the expectation is that this is an exporter to be done during edit time, or maybe at runtime but as an export task where the user don't expect to still interact with the application while it's happening, then I think it's fine to take this added hit for determinism.

If there are expectations that this export needs to happen "instantly" or in async/background while the application maintains interactive speeds, then we might want to do some perf exploration to understand the impact.

I'm assuming expectations are that exporting an alembic file is an expensive operation, I'm also assuming that exporting an alembic file is a synchronous main-thread operation that scales linearly with scene complexity, so there is no expectations that the editor or a player will keep interactive frame-rates while the export is happening. If this all holds, then I think its find to take a perf hit in favor of better determinism, this is the correct trade-off.

But please let me know if any of my assumptions doesn't hold!

LeoUnity (approved):
Replied to the perf topic. I'm approving assuming that the assumptions I raised are true, but if they are not, let me know and we can discuss a path forward!

com.unity.formats.alembic/Runtime/Scripts/Exporter/AlembicExporter_impl.cs
mfe: Great questions :D 

Alembic exporters (both the Alembic Exporter component and the Alembic recorder) work similarly as Recorder: you set-up the exporter in Edit mode but the export itself is only happening in Play mode.

My understanding is most of the time, the Alembic Importer / Exporter is used for round-tripping during production, but using it in a build is a valid use case.

In both case, there is no expectation that the export will be realtime and it is indeed expected that the deepest/complexe the hierarchy is, the slower it will be. As a matter of fact, the cost added by the sort is small as compared to the FindObjectsByType or the BakeMesh already done.

But that's another topic and I agree with you that true determinism is worth the trade off.

That said, I pushed a tiny improv to remove come GC allocs. I'm not sure it is worth to bloat the code but, ya 😅

LeoUnity (approved):
Looks great! Ship it :)

